Issue #4, opened 2021-10-24
Should we add a support model to the raw PDFium bindings?


mara004:
I was thinking about whether we should  perhaps add a support model (i. e. custom objects and functions) on top of PDFium to make usage a bit more comfortable and more 'pythonic' than just the raw C-style interface.

First of all, there is the strange, but required `FPDF_InitLibrary()` call. Should we automatically do this in our `__init__.py` file?

Secondly, rendering a page is a bit laborious in terms of code, due to the necessity of ctypes casting and so on. I was thinking about whether we could add a simplified function `render_page()` that takes a PDFium document, a page index, rotation, scaling, and background colour parameters that returns the rasterised page as PIL image.

Another thing that annoys me is that one must always pass all arguments, and that it's not possible to use keyword arguments. A way to change this would be to create more custom objects and methods on top of PDFium, but this would be kind of a maintenance burden, and it might be less powerful than accessing PDFium directly.

PR #6, opened 2021-10-26
Add initial parts for a support model (cc #4)

mara004:
This sets up a basic support model on top of the lower-level PDFium API, to simplify common use cases:
* automatically call `FPDF_InitLibrary()`
* add a context manager `PdfContext()` for opening documents in a `with`-block
* add a function `render_page()` to easily rasterise a page of an `FPDF_DOCUMENT` to a `PIL.Image.Image`

### DIFF
--- a/src/pypdfium2/_helpers.py
+++ b/src/pypdfium2/_helpers.py
@@ -70,7 +70,7 @@ def render_page(
         rotation: int = 0,
         background_colour: int = 0xFFFFFFFF,
         render_annotations: bool = True,
-        optimise_mode: OptimiseMode = OptimiseMode.lcd_display,
+        optimise_mode: OptimiseMode = OptimiseMode.none,
     ) -> Image.Image:
     """
     Rasterise a PDF page using PDFium.


README.md
```diff
@@ -89,6 +90,34 @@ if __name__ == "__main__":
...
+    
+    with pdfium.PdfContext(filename) as pdf:
+        pil_image = pdfium.render_page(
+            pdf,
+            page_index = 0,
+            scale = 1,
+            rotation = 0,
+            background_colour = 0xFFFFFFFF,
+            render_annotations = True,
+            optimise_mode = pdfium.OptimiseMode.lcd_display,
```
adam-huganir, line 115: Is there a reason you are using FPDF_LCD_TEXT as a default? In my opinion it really messes up scaling, lemme upload a screenshot, I have an example from when I was working with this last month.

Normal: https://github.com/adam-huganir/misc/blob/main/default.png
With LCD: https://github.com/adam-huganir/misc/blob/main/with_lcd.png

The source document is black and white

mara004: > Is there a reason you are using FPDF_LCD_TEXT as a default?

I just took this over from the original pypdfium example. I have changed the default to `OptimiseMode.none`

mara004: @adam-huganir I've searched a bit about this topic. What FPDF_LCD_TEXT does is called [subpixel rendering](https://en.wikipedia.org/wiki/Subpixel_rendering) and it *is* considered to improve viewing quality on LCD displays.

adam-huganir: @mara004  My use cases come mostly from document processing (feeding into vision models or OCR) where it can lead to issues, but for displaying for users it is definitely better on the eyes. 

I am fine with either being the default as long as it is an option to turn them on or off

PR #7, opened 2021-11-10
Add a command-line interface

mara004:


src/pypdfium2/__main__.py
```diff
@@ -0,0 +1,231 @@
...
+
+def main():
+    
+    args = parse_args()
+    
+    if args.version:
+        print(f"PyPDFium2 {pdfium.__version__}", f"PDFium {pdfium.__pdfium_version__}", sep='\n')
+        sys.exit()
+    
+    assert args.pdffile is not None
```
adam-huganir: `-O` as a python argument will disable assert statements, so we should make this into an explicit `ArgumentError`

src/pypdfium2/__main__.py
```diff
@@ -0,0 +1,231 @@
...
+from os.path import (
+    join,
+    basename,
+    splitext
+)
+
+
+def rotation_type(string):
+    rotation = int(string)
+    assert rotation in (0, 90, 180, 270)
```
adam-huganir: assert -> raise ValueError

src/pypdfium2/__main__.py
mara004: I've changed this to ValueError - didn't figure out how to use `argparse.ArgumentError`.

src/pypdfium2/__main__.py
mara004: done

PR #12, opened 2021-11-30
Project organization

adam-huganir:
Ok, so the main change here is to use [poetry](https://python-poetry.org/docs/) for environment management and dependency management, and to update to uyse pyproject.toml and PEP 518. Not 100 % there yet, since it is not fully integrated with the wheel build system, but that will be one of the next things I tackle. 

I also ran a bunch of formatting and linting on the project files (black, isort, and flake8) so most if not all of the python code changes here are just formatting/removing unused variables/renaming variables that shadowed globals.

### DIFF
--- a/src/pypdfium2/__main__.py
+++ b/src/pypdfium2/__main__.py
@@ -56,8 +56,7 @@ def pagetext_type(value):
             page_indices.extend(pages)
 
         else:
-
-            page_indices.append(int(page_or_range))
+            page_indices.append(int(page_or_range) - 1)
 
     return page_indices
 
--- a/src/pypdfium2/_helpers.py
+++ b/src/pypdfium2/_helpers.py
@@ -8,9 +8,9 @@ from typing import Optional
 
 from PIL import Image
 
-import pypdfium2 as pdfium
 from pypdfium2._constants import *
 from pypdfium2._exceptions import *
+from pypdfium2 import _pypdfium as pdfium
 
 
 class PdfContext:
@@ -99,7 +99,8 @@ def render_page(
     Parameters:
 
         pdf:
-            A PDFium document.
+
+            A PDFium document (can be obtained with :class:`PdfContext`).
 
         page_index:
             Zero-based index of the page to render.
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,46 @@
+# SPDX-FileCopyrightText: 2021 Adam Huganir <adam@huganir.com>
+# SPDX-License-Identifier: CC-BY-4.0 
+
+release: clean update-binaries build-all
+	check-wheel-contents dist/*
+
+build-all: build-darwin-arm64 build-darwin-x64 build-linux-arm32 build-linux-arm64 \
+			build-win-x64 build-win-x86 build-linux-x64
+
+build-darwin-x64:
+	python3 setup_darwin_x64.py   bdist_wheel
+
+build-darwin-arm64:
+	python3 setup_darwin_arm64.py bdist_wheel
+
+build-linux-x64:
+	python3 setup_linux_x64.py    bdist_wheel
+
+build-linux-arm64:
+	python3 setup_linux_arm64.py  bdist_wheel
+
+build-linux-arm32:
+	python3 setup_linux_arm32.py  bdist_wheel
+
+build-win-x64:
+	python3 setup_windows_x64.py  bdist_wheel
+
+build-win-x86:
+	python3 setup_windows_x86.py  bdist_wheel
+
+update-binaries:
+	python3 update.py
+
+clean:
+	-rm -r dist
+	-rm -r build
+	-rm -r data/*
+	-touch data/.gitkeep
+	-rm -r src/pypdfium2.egg-info/
+	-rm -r __pycache__
+	-rm -r src/pypdfium2/__pycache__
+	-unlink src/pypdfium2/_pypdfium.py
+	-unlink src/pypdfium2/pdfium
+	-unlink src/pypdfium2/pdfium.dll
+	-unlink src/pypdfium2/pdfium.dylib
+


.gitignore
```diff
@@ -1,19 +1,49 @@
...
+env/
+venv/
+ENV/
+env.bak/
+venv.bak/
+
+# Mypy
+.mypy_cache/
+.dmypy.json
+dmypy.json
```
mara004, line 45: Why `.dmypy.json` and `dmypy.json`? When/where does mypy generate these files?

.reuse/dep5
```diff
@@ -1,14 +0,0 @@
...
-
-# Sample paragraph, commented out:
-#
-# Files: src/*
-# Copyright: $YEAR $NAME <$CONTACT>
-# License: ...
-
-Files: pypdfium2.kdev4 .kdev4/*
-Copyright: 2021 geisserml <geisserml@gmail.com>
-License: CC-BY-4.0
```
mara004, line 14: I'd prefer to keep the `dep5` file as it's for the `reuse` tool.

.kdev4/pypdfium2.kdev4
```diff
@@ -1,53 +0,0 @@
-[Buildset]
```
mara004, line 1: You're right IDE projects don't usually belong into the repository, but this was nice for my convenience so I can delete and re-clone the local repository without loosing anything, and it helps other people using kdevelop to get the correct configuration right away...

poetry.lock
```diff
@@ -0,0 +1,882 @@
+[[package]]
```
mara004, line 1: Is this file copy-pasted? It contains lots of stuff we don't use, like `babel`, `alabaster`, `colorama`, and so on.
I don't think we can merge the file in this state.

pyproject.toml
```diff
@@ -0,0 +1,45 @@
...
+    "Intended Audience :: Developers",
+    "Intended Audience :: Information Technology",
+    "Intended Audience :: Education",
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: Implementation :: CPython",
+    "Topic :: Multimedia :: Graphics",
+    "Topic :: Software Development :: Libraries"]
+keywords = ["Python", "PDF", "PDFium"]
+
```
mara004, line 21: Hmm, this is kind of duplicated from `setup.cfg` - Why doesn't poetry read it from there?

adam-huganir: Yeah, I grabbed it directly from there, it is just a matter of PEP 518 has standardized using a pyproject.toml for this type of stuff over setup.cfg. Unfortunately, not everything has switched to pyproject.toml. The main reason for it now is dependency management to make development by multiple people always yield the same results/minimize issues with potential dependency conflicts

examples/render_supportmodel.py
```diff
@@ -1,27 +1,27 @@
...
-            rotation = 0,
-            background_colour = 0xFFFFFFFF,
-            render_annotations = True,
-            optimise_mode = pdfium.OptimiseMode.none,
+            page_index=0,
+            scale=1,
+            rotation=0,
+            background_colour=0xFFFFFFFF,
+            render_annotations=True,
+            optimise_mode=pdfium.OptimiseMode.none,
```
mara004, line 24: I'm not very happy about this reformatting, tbh.

setup.cfg
```diff
@@ -7,7 +7,7 @@ version = attr: pypdfium2._version.V_PYPDFIUM2
 description = Python bindings to PDFium
 long_description = file: README.md
 long_description_content_type = text/markdown
-author = pypdfium2-team
+author = geisserml
```
mara004, line 10: I'd prefer to keep `pypdfium2-team` here, as I'm not the sole author.

setup.cfg
```diff
@@ -29,7 +29,7 @@ keywords = Python, PDF, PDFium
 packages = find:
 package_dir =
     =src
-python_requires = >=3.6
+python_requires = >=3.5
```
mara004, line 32: Actually we require python 3.7, at least on windows. That's because we're using `str.isascii()` in the CLI, and this method only got added with 3.7
In any case, we're using f-strings at some places, so 3.5 won't work.

examples/render_supportmodel.py
adam-huganir: We don't have to use `black`, it just seems to be the most common nowadays so I am used to it. The dev even says in the header on their splash page: `The uncompromising code formatter`, and it is not an understatement

We can switch to another one if you want, or even just not worry about standardizing, it's up to you

src/pypdfium2/__main__.py
```diff
@@ -23,151 +20,159 @@ def rotation_type(string):
...
-            
+                pages = [i for i in range(start, end - 1, -1)]
+
             page_indices.extend(pages)
-        
+
         else:
-            
+
             page_indices.append(int(page_or_range))
```
mara004: `-1` from the other branch needed here

src/pypdfium2/_constants.py
```diff
@@ -4,8 +4,9 @@
 from enum import Enum
 
 
-class OptimiseMode (Enum):
```
mara004, line 7: I actually like a space before the inheritance brackets...

setup.cfg
adam-huganir: weird, I don't even remember changing this, 🤷

src/pypdfium2/_constants.py
```diff
@@ -4,8 +4,9 @@
...
+class OptimiseMode(Enum):
     """ How to optimise page rendering """
-    none = 0          #: Do not use any optimisations
-    lcd_display = 1   #: Optimise for LCD displays
-    printing = 2      #: Optimise for printing
```
mara004, line 11: I prefer the block formatting - it's easier to read

src/pypdfium2/_helpers.py
```diff
@@ -82,84 +84,86 @@ def _translate_rotation(rotation: int):
...
 def render_page(
-        pdf: pdfium.FPDF_DOCUMENT,
-        page_index: int,
-        *,
-        scale: float = 1,
-        rotation: int = 0,
-        background_colour: int = 0xFFFFFFFF,
-        render_annotations: bool = True,
-        optimise_mode: OptimiseMode = OptimiseMode.none,
-    ) -> Image.Image:
```
mara004, line 93: nope, I want to keep this indentation level

update.py
```diff
@@ -5,176 +5,181 @@
...
+    "https://github.com/bblanchon/pdfium-binaries/releases/download/chromium%2F"
+)
 ReleaseFiles = {
-    'pdfium-mac-arm64.tgz'    : 'darwin-arm64',
-    'pdfium-mac-x64.tgz'      : 'darwin-x64',
-    'pdfium-linux-arm.tgz'    : 'linux-arm32',
-    'pdfium-linux-arm64.tgz'  : 'linux-arm64',
-    'pdfium-linux-x64.tgz'    : 'linux-x64',
-    'pdfium-win-x64.tgz'      : 'windows-x64',
-    'pdfium-win-x86.tgz'      : 'windows-x86',
```
mara004, line 31: again preferring block syntax.

setup.cfg
mara004: Oh, I know how this happend. Your branch might be based on a slighly older state of the repository. That was a recent change.

examples/render_supportmodel.py
mara004: Of course it would be nice to have some kind of consistency in code formatting, but what `black` did is rather decreasing readability IMHO.

adam-huganir: autopep8 is one that allows for much more flexibility on what to allow and what to exclude, so we could use that one instead and disable the options that clash with these

mara004: We can certainly look into this at some point, but code formatting is currently not a priority for me

PR #15, opened 2021-11-30
added dev makefile for basic scripts

adam-huganir:
just the makefile

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,5 @@
-
+# SPDX-FileCopyrightText: 2021 Adam Huganir <adam@huganir.com>
+# SPDX-License-Identifier: CC-BY-4.0 
 
 release: clean update-binaries build-all
 	check-wheel-contents dist/*


Makefile
```diff
@@ -0,0 +1,45 @@
+
```
mara004:
```
# SPDX-FileCopyrightText: 2021 Adam Huganir <adam@huganir.com>
# SPDX-License-Identifier: CC-BY-4.0 
```

Makefile
```diff
@@ -0,0 +1,45 @@
...
+
+release: clean update-binaries build-all
+	check-wheel-contents dist/*
+
+build-all: build-darwin-arm64 build-darwin-x64 build-linux-arm32 build-linux-arm64 \
```
mara004, line 7: Why does this have a backslash at the end, but  line 3 doesn't?

adam-huganir: So the first one is saying do these tasks first, and then run `check-wheel-contents` in `sh` (so the first line is the prerequisites and the second is the first command0).

the build-all is just a line continuation, the first line since it is so long, so make interprets it as one lone line

mara004: Ah, yes. Okay.

Makefile
adam-huganir: done

PR #17, opened 2021-11-30
Add preliminary tests for the command-line interface

adam-huganir:
fixed weird merging issue with previous branch

### DIFF
--- a/tests/__init__.py
+++ b/tests/__init__.py
@@ -1,0 +1,3 @@
+# SPDX-FileCopyrightText: 2021 Adam Huganir <adam@huganir.com>
+# SPDX-FileCopyrightText: 2021 geisserml <geisserml@gmail.com>
+# SPDX-License-Identifier: Apache-2.0


tests/test___main__.py
```diff
@@ -0,0 +1,83 @@
...
+    "test_input,expected",
+    [
+        (["python"], _base_parsed_args),
+        (["python", "--pages=1,2"], {**_base_parsed_args, "pages": [1, 2]}),
+        (["python", "--scale=3.14"], {**_base_parsed_args, "scale": 3.14}),
+    ],
+)
+def test_parse_args(test_input, expected):
+    # assert vars(parse_args(test_input)) == expected
+    assert True
```
mara004: What is the problem with this test?

tests/test___main__.py
```diff
@@ -0,0 +1,83 @@
+# SPDX-FileCopyrightText: 2021 geisserml <geisserml@gmail.com>
```
mara004, line 2: You're the author, not me ;)

mara004: Changed it now

tests/test___main__.py
mara004: modified/fixed

PR #28, opened 2021-12-12
Add a support model for reading the table of contents

mara004:
Seems to work well in general and even appears to be quite performant, but two points are missing:

* Determining the state of an outline item (opened/collapsed). PDFium/Chromium do not support this as far as I can see. That's why I filed a [feature request](https://bugs.chromium.org/p/pdfium/issues/detail?id=1758) upstream.
* Test integration. I have tested manually using the CLI, but it's not part of our test suite yet, because I haven't found a suitable document that is both small and shareable.

### DIFF
--- a/src/pypdfium2/__main__.py
+++ b/src/pypdfium2/__main__.py
@@ -133,6 +133,11 @@ def parse_args(args=sys.argv[1:]):
         help = "The number of processes to use for rendering. Defaults to the number of CPU cores."
     )
     parser.add_argument(
+        '--show-toc',
+        action = 'store_true',
+        help = "Print the table of contents of a PDF document."
+    )
+    parser.add_argument(
         '--version', '-v',
         action = 'store_true',
         help = "Show the program version and exit."
@@ -151,6 +156,12 @@ def main():
     
     if args.pdffile is None:
         raise ValueError("An input file is required.")
+    
+    if args.show_toc:
+        with helpers.PdfContext(abspath(args.pdffile)) as pdf:
+            toc = helpers.get_toc(pdf)
+            helpers.print_toc(toc)
+        sys.exit()
     
     if args.prefix is None:
         prefix = splitext(basename(args.pdffile))[0] + '_'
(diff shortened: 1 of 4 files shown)


mara004:
While the current implementation seems to work, there are some points where I'm pretty uncertain:

* Length of the string buffer for the title: Currently, I'm calling `FPDFBookmark_GetTitle()` twice, to get the buffer size first.

  ```python3
    t_buflen = pdfium.FPDFBookmark_GetTitle(bookmark, None, 0)
    t_buffer = ctypes.create_string_buffer(t_buflen)
    pdfium.FPDFBookmark_GetTitle(bookmark, t_buffer, t_buflen)
    title = t_buffer.raw[:t_buflen].decode('utf-16-le')[:-1]
  ```
  However, Kuafu always assumes a buffer length of `4096`:
  https://github.com/YinlinHu/kuafu/blob/dd9854eb51e7dcff9d91462cf7151586f3d52697/kuafu/pdfworker.py#L194  
  ```python3
        buflen = 4096
        titleBuf = ctypes.create_string_buffer(buflen)
        titleByteLen = PDFIUM.FPDFBookmark_GetTitle(bookmark_node, titleBuf, buflen)
        assert(titleByteLen <= buflen)
        titleStr = titleBuf.raw[:titleByteLen].decode('utf-16-le')[:-1]
  ```

* Converting the viewport position from C to a Python type: At the moment, I have created a custom array type:
  ```python3
    class ArrayFSFloat4 (ctypes.Array):
        _type_ = pdfium.FS_FLOAT
        _length_ = 4
  ```
  ```python3
    n_params = ctypes.c_ulong()
    view_pos = ArrayFSFloat4()
    view_mode = pdfium.FPDFDest_GetView(dest, n_params, view_pos)
    n_params = n_params.value
    view_pos = list(view_pos)[:n_params]
    view_mode = _translate_viewmode(view_mode)
  ```
   Is this okay or should I do it differently? I have literally zero experience with ctypes, that's why I'm not sure..,.

mara004:
Whoops, that merge was accidental. I pushed to the wrong upstream. That happens when you're working at night...
However, I think this was in a state that it can be kept in main. If you have an suggestions or improvements, please let me know!

Issue #33, opened 2021-12-18
Support building PDFium on non-standard architectures


mara004:
While testing sourcebuild on Linux armv7l, I noticed that the PDFium build process relies on several pre-built binaries from depot-tools that are only available for the most common architectures.
Therefore, we will need to adapt our build script to use the `gn` and `ninja` executables provided by the system, if available. On Debian, the relevant packages are named [`generate-ninja`](https://packages.debian.org/bullseye/generate-ninja) and [`ninja-build`](https://packages.debian.org/bullseye/ninja-build).

mara004:
Unfortunately, PDFium build tools also ship their own llvm-based compiler toolchain, rather than using the system-provided clang. Moreover, it seems like they are using build flags that are not supported on older versions of clang. This makes building on non-standard architectures yet a lot more complicated.

mara004:
It still won't be easily possible to bulid PDFium on non-standard architectures, but I don't know what else to do on my side for now.

mara004:
I've put some more analysis into this and could trace the problem down to the following places:

Binaries need to be replaced with symlinks to the corresponding system-provided executables:
```
pdfium/third_party/llvm-build/Release+Asserts/bin/
```

Configuration needs to be patched to remove some problematic clang parameters:
```
pdfium/build/config/compiler/BUILD.gn
pdfium/build/config/clang/BUILD.gn
```

mara004:
I managed to get compiling work locally now, but linking still fails with a cryptic error message reporting that ldd does not accept a `-z` option. However, judging from the printed command, `-z` is not even passed.

mara004:
I've got it working now with a more rigid patch...

mara004:
Some work has been done in the [`nativebuild`](https://github.com/pypdfium2-team/pypdfium2/tree/nativebuild) branch that might enable building with system-provided compilers/linkers. However, it seems that PDFium is completely centered towards cross compilation and not made for native compiling on 'special' architectures. On my arm32 machine, it just fails with the compiler complaining about the sources.

PR #37, opened 2021-12-19
Proposal: Implicit sourcebuild as fallback setup strategy

mara004:
The advantage is that this would enable the possibility of publishing an sdist package to PyPI so that it will be attempted to build PDFium from source on platforms where no wheel is available.
However, performing an implicit sourcebuild without the user consciously opting in might not always be desirable, especially considering that setuptools hides the output of the build script, and given that PDFium unfortunately won't easily build on non-standard architectures (see #33).
Nevertheless, I believe sdist would be a good idea. What do you think?


mara004:
I'm merging this now for today's build

PR #42, opened 2021-12-22
Enable building with system-provided compiler/linker

mara004:
Some changes in the patch might be unnecessary, but it is hard to figure out which flags can be kept...


mara004:
I was able to greatly shorten the patch by setting certain config flags, and most importantly by fixing the `ldd`/`lld` blunder.

PR #44, opened 2022-01-02
Integrate support for Windows ARM64

mara004:
I'm not sure whether PyPI support `win_arm64` wheels at all, but we'll see tomorrow. The code of [`sysconfig.get_platform()`](https://github.com/python/cpython/blob/3.10/Lib/sysconfig.py#L691) suggests that it might work.

### DIFF
--- a/release.sh
+++ b/release.sh
@@ -5,4 +5,4 @@
 bash clean.sh
 python3 update.py
 bash setup_all.sh
-check-wheel-contents dist/*
+check-wheel-contents dist/*.whl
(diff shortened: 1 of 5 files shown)


mara004:
If it's really not supported yet, I think I'll keep the `setup_windows_arm64.py` file and only comment the platform out in `update.py` and `setup_all.sh`, so that Windows ARM64 users can build and install the package locally.

mara004:
Note to self for tomorrow: Remember to upload to TestPyPI first!

mara004:
Just uploaded to TestPyPI - it works!

PR #55, opened 2022-01-28
Completely restructure the setup strategy

mara004:
Previously, running `setup.py` always triggered a source build, on behalf of platforms where no wheel is available.
This MR changes the setup strategy in a fundamental way. Now `setup.py` determines the current platform and uses pre-built binaries if available, with source build only as fallback. This is quite advantageous:
* Users who wish to install PyPDFium2 from git main can do so by just downloading the repository and running
  `pip3 install . -v`, as this now won't trigger a time-consuming and uncertain source build anymore unless it's really necessary.
* Re-installing PyPDFium2 during development is simplified a lot.
* ReadTheDocs can now be configured in a saner way. We are no longer bound to the current release on PyPI, but can just check the option `Install Project` and remove `pypdfium2` from `docs/requirements.txt`.

On the way of restructuring the setup, I also addressed another issue that I was concerned about since a longer time:
The setup code was ran twice, because pip has to execute the file repeatedly with different commands. This can be quite a slowdown if a lot of tasks have to be done, such as dependency installation and all the build commands. Luckily, PDFium's build system has been intelligent enough to detect that it doesn't need to recompile with the second run.
I now fixed this using a text file `data/setup_status.txt` to track the current setup state. By default, it contains the string `InitialState`. On the first setup run, all tasks that only need to be done once are performed, and status is updated to `PreSetupDone`. In the following run, `PreSetupDone` is detected and the tasks are skipped.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,7 @@
 # SPDX-License-Identifier: CC-BY-4.0
 
 data/
-!data/.gitkeep
+!data/setup_status.txt
 build/
 dist/
 docs/build/*
--- a/.gitignore
+++ b/.gitignore
@@ -2,7 +2,6 @@
 # SPDX-License-Identifier: CC-BY-4.0
 
 data/
-!data/setup_status.txt
 build/
 dist/
 docs/build/*
--- a/.setup_status.txt
+++ b/.setup_status.txt
@@ -1,0 +1,1 @@
+InitialState
--- a/.setup_status.txt
+++ b/.setup_status.txt
@@ -1,1 +1,0 @@
-InitialState
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,8 +29,8 @@
   build, on behalf of platforms for which no wheel is available. With this release, the code
   was changed to detect the current platform and use pre-compiled binaries if available, with
   source build only as fallback.
-- The version file is now always read literally on setup, which makes it a lot less prone to
-  errors.
+- On setup, the version file is now always read literally (i. e. without importing the module),
+  which makes it a lot less prone to errors.
 - Modernised the update script code that reads and writes the version file.
 - Updated the Makefile.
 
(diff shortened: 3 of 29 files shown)


mara004:
Note to self: remember squashing

mara004:
Remaining Tasks:
* Update changelog

mara004:
I just tried a source build using setup.py by commenting out the platform-specific binary handler.
It currently fails for reasons that look fairly obscure. I'll hopefully be able to investigate this by tomorrow.

mara004:
Oh, found the spot. It was pretty trivial after all, just a bit hard to find. I was missing the `argv` parameter for `build_pdfium.parse_args()`! It should just be an empty list, otherwise the arguments of setuptools will be used by mistake.

mara004:
There's one thing that strongly confuses me: the output order of the log.
Downloading PDFium binray packages is to be run *before* bindings creation. In the output, however, It's the other way round:
```
  INFO: Status: Preprocessing /tmp/tmpkjcs2e4i.h
  INFO: Status: gcc -E -U __GNUC__ -dD "-Dinline=" "-D__inline__=" "-D__extension__=" "-D__const=const" "-D__asm__(x)=" "-D__asm(x)=" "-DCTYPESGEN=1" "/tmp/tmpkjcs2e4i.h"
  INFO: Status: Parsing /tmp/tmpkjcs2e4i.h
  ERROR: /usr/include/x86_64-linux-gnu/sys/cdefs.h:245: Syntax error at '\n'
  INFO: Status: Processing description list.
  INFO: Status: Writing to ~/projects/pypdfium2/data/linux-x64/_pypdfium.py.
  INFO: Status: Wrapping complete.
  No new PDFium build - will re-create bindings without incrementing version
  Downloading https://github.com/bblanchon/pdfium-binaries/releases/download/chromium%2F4849/pdfium-linux-x64.tgz -> ~/projects/pypdfium2/data/linux-x64/pdfium-linux-x64.tgz
```

I don't know what causes this strange output order, but it looks like the order of execution is correct anyway, otherwise bindings generation couldn't work at all.

mara004:
This now looks quite ready for merging to me

PR #58, opened 2022-02-05
Implement workaround for non-ascii files on Windows (see #57)

mara004:


mara004:
#### Tasks
* Add individual tests for each method. In general, it might be best to split up the helpers tests into multiple files as the support model is constantly growing.

PR #61, opened 2022-02-08
Experiment with rendering using custom pixel formats

mara004:
Directly initialise the PDFium bitmap with the right colour format, rather than converting afterwards using PIL.
This should improve performance for rendering without alpha channel, especially for greyscale.


mara004:
It's a pity that PDFium doesn't seem to offer a pixel format for greyscale with alpha (`LA` colourspace). We still have to use RGBA for that...

PR #68, opened 2022-02-17
Initial steps for making pypdfium2 setup compliant with PEP517

mara004:


mara004:
It's a small wonder, but I think this might actually be a bit better than before. I'm starting to like PEP 517.
I'll merge this in and see how it goes.

mara004:
Hmm, the latest docs build is broken now. From looking at the logs, it seems that readthedocs installs the project using `setup.py install`, rather than `pip3 install .`
We might have to call getdeps in setup.py anyway to achieve compatibility with `setup.py install` again.
Or we could finally create a `.readthedocs.yaml` to configure the install the way we'd like it...

PR #72, opened 2022-02-21
Add experimental autodocgen utility

mara004:
Alternative approach for #70


mara004:
It's a nice proof of concept that was interesting to write, but I'm afraid it's not very practical.
There are no docstrings, only the plain attribute names. Not even method parameters or their types are included.
I think we'd be better of building the PDFium doxygen documentation locally...

PR #73, opened 2022-02-22
make build/update callable without argparse

mara004:


mara004:
Maybe not perfect but I think it's a step forward.

PR #75, opened 2022-02-22
Start removing type hints

mara004:


mara004:
I think I don't want to do this for now -- it would only bring compatibility with Python 3.4 or maybe 3.3, but I couldn't test it anyway since pyenv fails to build Python 3.4 (depends on outdated OpenSSL lib).

PR #78, opened 2022-02-25
Convert `PlatformDirs` to `PlatformNames`

mara004:


mara004:
Hard to tell whether this would be an improvement. I'm not particularly happy with the current situation, but this doesn't really seem better yet.

mara004:
I like the change.

PR #80, opened 2022-03-01
Start restructuring renderer components to make PIL optional

mara004:


mara004:
Amazing. I've been longing for this change since quite some time.

mara004:
Concerning the deprecated functions, I'll see whether we will keep them for the next release or directly remove the lot. Probably the latter, so we can have all API-breaking changes at once.

PR #81, opened 2022-03-01
Remove Windows workaround for non-ascii file paths

mara004:
PDFium team claim to have fixed these issues.


mara004:
Just confirmed on Windows 8.1 that it really works.

Issue #83, opened 2022-03-02
`platform_setup` imports not working when scripts are invoked directly?


mara004:
Whilst testing on a different device, I encountered a behaviour that I had not experienced on my main workplace yet:
When the scripts are invoked directly, the imports from `platform_setup` fail. Actually, this is quite logical, as the top-level directory is not in `PYTHONPATH`. I'll have to investigate why it works on the other device.
(We are already using importlib in `setup.py` to load `platform_setup` manually, so this part should be unaffected.)

PR #84, opened 2022-03-02
Add boilerplate code to ensure scripts can be run directly

mara004:
Looks terrible, but it's the only realistic solution for #83 that comes to my mind...


mara004:
Really outrageous in terms of code style, but I think in this case it's just more important to ensure the scripts run everywhere.

PR #94, opened 2022-03-11
Build improvements

mara004:
* Improved PDFium download/update
* Added option to check out a certain PDFium revision
* Renamed depot tools patch directory to match the repository name
* Added capabilities to skip unnecessary PDFium dependencies (huge performance improvement for first-time downloading)


mara004:
As of b3f60dc, if I run `time python3 platform_setup/build_pdfium.py -ul` with the repository already being downloaded, I get the following result:
```
real    2m13,925s
user    7m9,597s
sys     0m24,341s
```

Issue #103, opened 2022-04-30
Add musllinux to setup


mara004:
Upstream pdfium-binaries added build support for musllinux quite recently. We should thus integrate it into our setup code, as per [`add_platform.md`](https://github.com/pypdfium2-team/pypdfium2/blob/main/docs/markdown/add_platform.md). It might take a while until I have some time to work on this.

PR #104, opened 2022-04-30
Prepare musllinux integration (CC #103)

mara004:
Not sure how to properly recognise musl host systems in `setup.py` (see https://github.com/python/cpython/issues/87414).


setupsrc/pl_setup/setup_base.py
```diff
@@ -49,6 +49,9 @@ def _copy_bindings(pl_name):
 def _get_linux_tag(arch):
     return 'manylinux_2_17_%s.manylinux2014_%s' % (arch, arch)
 
+def _get_musllinux_tag(arch):
+    return 'musllinux_1_2_%s' % (arch)
```
mara004, line 53: `1_1` might work as well - not sure about this.

PR #105, opened 2022-05-06
Add helper model for text insertion, supporting Asian scripts

mara004:
Many thanks to Anurag Bansal who sent me his text insertion script and kindly consented that I may incorporate it into pypdfium2 as a new support model.


src/pypdfium2/_helpers/textinserter.py
```diff
@@ -0,0 +1,96 @@
...
+        text,
+        pos_x,
+        pos_y,
+        font_path,
+        font_size,
+        font_type,
+        font_is_cid,
+    ):
+    """
+    Insert text into a PDF page at a specified position. This function supports Asian scripts such as Hindi.
```
banagg: The comment can probably be - "Insert text into a PDF page at a specified position using the language's ligature (https://en.wikipedia.org/wiki/Ligature_(writing))", as it's not just re-arrangement which is happening since these languages usually combine two characters to form a entirely new character (usually called as glyphs, and identified using cid - character id) which in themselves do not have a unicode value assigned to them.

Maybe even better explanation can be added then the one I provided.

src/pypdfium2/_helpers/textinserter.py
```diff
@@ -0,0 +1,96 @@
...
+        font_type (int):
+            State the type of the given font (``FPDF_FONT_TYPE1`` or ``FPDF_FONT_TRUETYPE``).
+        font_is_cid (bool):
+            State whether the given font is a CID font.
+    """
+    
+    if not have_harfbuzz:
+        raise RuntimeError("Function insert_text() requires uharfbuzz to be installed.")
+    
+    with open(font_path, "rb") as fh:
```
banagg: Can an option be added using which the font is not required to be loaded again and again (considering that someone would want to insert multiple lines of text). Something of an optional argument which says that if a font is provided then when inserting a text, the font loading is not required.

src/pypdfium2/_helpers/textinserter.py
```diff
@@ -0,0 +1,96 @@
...
+        font_data = fh.read()
+    
+    page = open_page(pdf, page_index)
+    
+    hb_blob = harfbuzz.Blob.from_file_path(font_path)
+    hb_face = harfbuzz.Face(hb_blob)
+    hb_font = harfbuzz.Font(hb_face)
+    hb_fontscale = hb_font.scale[0]
```
banagg: Also, maybe the harfbuzz stuff can be added in a utility function as well, considering that the hb_fontscale variable would always have one value for a given font. I understand the font loading would not take much time, but maybe adding the optional param would optimize this code for making large PDFs.

src/pypdfium2/_helpers/textinserter.py
```diff
@@ -0,0 +1,96 @@
...
+    
+    pdf_font = pdfium.FPDFText_LoadFont(
+        pdf,
+        ctypes.cast(font_data, ctypes.POINTER(ctypes.c_uint8)),
+        len(font_data),
+        font_type,
+        font_is_cid,
+    )
+    
+    hb_buffer = harfbuzz.Buffer()
```
banagg: I agree that this is the logic I provided. Here I mentioned to create the hb_buffer object for each text typesetting info we add. Honestly, it was a bit of lazy code writing on my end, and there should probably be something better available using which the buffer object is not required to be created again and again.

banagg (commented):
The text_inserter code would indeed work. I mainly have a few issues with the potential optimizations which can be done regarding font loading as mentioned in the individual comments.

src/pypdfium2/_helpers/textinserter.py
mara004: That's a good point. What about requiring the caller to pass `font_data`?

mara004:
Thanks very much for the review. It's good you mention the performance aspects, which I hadn't really thought about yet. I guess we can move any font-specific computation into an own function that returns a data object, and require this to be passed to the function.

src/pypdfium2/_helpers/textinserter.py
mara004: Yes, that sounds better. I'll add it.

mara004:
I have changed the API so that it should now support multiple text insertions in a more performant way. Can you please review again?

src/pypdfium2/_helpers/textinserter.py
mara004: I now wrote `using the writing system's ligature`, as the code only gets the characters and doesn't know which language it is.

src/pypdfium2/_helpers/textinserter.py
mara004: Done with implementation of `open_pdffont()`.

src/pypdfium2/_helpers/textinserter.py
mara004: Done with implementation of `open_pdffont()`.

banagg (approved):
This version looks great. Thanks for the update.

src/pypdfium2/_helpers/textinserter.py
mara004: I'm not exactly sure what you mean? Do you think the buffer object should be initialised in a different scope?

mara004:
That's good to hear!

src/pypdfium2/_helpers/textinserter.py
banagg: I meant as of now, we are creating the buffer each and every time the create text function is loaded. And the solution does make sense considering that when we try to fetch the glyph data for more text which can be inserted into the hb buffer, the solution does not work any more and returns empty blocks (you can try this behaviour).

Now if hb_buffer creation in itself is not a heavy task, then I don't think that any update is required, but if it is something which can create problems in large PDFs then this probably needs some investigation.

Obviously, I am using this version on my end, as I do not have a better alternative for this on my end.

mara004: OK, thanks for the clarification.

PR #113, opened 2022-06-05
Experimental direct RGB(A) rendering

mara004:
This change would make PDFium render RGB(A) directly, instead of BGR(A), using the `pdfium.FPDF_REVERSE_BYTE_ORDER` flag. Might be advantageous as RGB(A) is more common, so callers who directly work with the bytes could use them as-is and would not need to change the byte order. Perhaps it might also make constructing the PIL image slightly faster (need to check).

However, this adds additional complexity: The aforementioned flag doesn't work well with `FPDFBitmap_Gray`, as some parts are missing in the result (probably due to a PDFium bug). It doesn't really matter because that's only one channel anyway, but it means we need to handle background colour separately. RGB(A) bitmaps would need ABGR, and L bitmaps ARGB, then. Rather confusing.

### DIFF
--- a/src/pypdfium2/_helpers/_utils.py
+++ b/src/pypdfium2/_helpers/_utils.py
@@ -4,9 +4,9 @@
 import pypdfium2._pypdfium as pdfium
 
 
-def colour_tohex(r, g, b, a=255):
+def colour_tohex(r, g, b, a=255, greyscale=False):
     """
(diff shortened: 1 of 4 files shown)


mara004:
The benchmark says both RGB and BGR have about the same performance for use with PIL.

mara004:
According to my benchmark, it looks `pdfium.FPDF_REVERSE_BYTE_ORDER` actually makes rendering to bytes a bit slower.

PR #115, opened 2022-06-09
Start experimenting with special methods

mara004:


mara004:
I'm not sure if this is going the right way. I have selectively ported some important improvements to main, but am yet undecided about the special methods.

PR #122, opened 2022-07-30
Rewrite non-RTD documentation and improve code quality on the way

mara004:


mara004:
It's not completely finished yet but I'll do the remaining sections in a different PR.

PR #130, opened 2022-09-01
Move form environment from page rendering to document level

mara004:
Initially troubled by a segmentation fault, which appears to have been caused be the (undocumented) requirement that the `FPDF_FORMFILLINFO` object needs to stay alive until the exit function is called.


mara004:
A member of PDFium team has confirmed my guess regarding object lifetime and will take care of PDFium's documentation.

PR #140, opened 2022-09-18
Greatly improve integration of rendering converters

mara004:
Provide standardised means to plug in converters.


mara004:
Note to self: remember to update the team readme and the stackoverflow post accordingly

mara004:
I think this is finally ready for being merged.

PR #142, opened 2022-09-24
Automatically close objects via `__del__`

mara004:


tests/helpers/test_opener.py
```diff
@@ -188,8 +188,6 @@ def test_open_new():
 
 
 def test_open_invalid():
-    with pytest.raises(TypeError, match=re.escape("__init__() missing 1 required positional argument: 'input_data'")):
-        pdf = pdfium.PdfDocument()
```
mara004, line 192: Removed this test case because it doesn't check anything pypdfium2-specific. This is common Python behaviour.

Thereby, we avoid an "unraisable exception" warning when running tests (`__del__` is still called, even if initialisation failed due to a parameter error handled by Python and the actual constructor code was never reached).

mara004: Hmm, theoretically, we could add some `hasattr()` check to suppress these warnings, but I don't think that's worth it, given that this only happens in case of serious API misuse.

PR #143, opened 2022-09-26
Rendering with shared memory

mara004:


mara004:
Hmm, considering the problem outlined [here](https://stackoverflow.com/questions/62748654/python-3-8-shared-memory-resource-tracker-producing-unexpected-warnings-at-appli), it might be better to pre-calculate the required buffer size, allocate the shared memory block in the main process, and resolve the shared memory name inside the `PdfPage.render_base()` function.

Also, I'm not quite sure if the allocator parameter is useful at all for what the documentation note says, because rendering will always happen in a different process for any GUI use case, so we'd always need shared memory for zero-copy rendering.

mara004:
Oh, I've got an idea for the inter-process callback: We could use the `.send()` method of the generator:
```python
# main process
for mem_size in renderer:
    shared_mem = # ... create shared memory block and get its name
    renderer.send(shared_mem)
   # ... work with pixel data in shared memory segment

# worker process
# ... calculate required size
shared_mem = yield mem_size  # request and obtain shared memory block
# ... render into shared memory segment
```

I've never tried something like that yet, so I'm not sure if it actually works, but it sounds promising at least.
This would be much easier than messing around with pipes/queues (ProcessPoolExecutor seems kind of ill-adapted for that in our case).

mara004:
Ah, mistake: we're sending to the wrong generator.
This change implies that `render_base()` would become a generator method, so that's the one we would need to send to.
However, as I take it we can't transfer the generator between worker and host process, that kind of brings back the previous problem.

mara004:
I also checked if we could use generators and `.send()`, but naturally, generators are not multiprocessing-compatible.

mara004:
Maybe using pipes and annotating the items with process IDs would work, though.

mara004:
Ah, I think I finally found a solution for the inter-process callback. But it's quite involved, with an additional request handler thread. I'll push this to a branch and make a snapshot in a PR, just to remember if we need it later.

I also thought about whether we can just pre-calculate size, but I don't think this will work out, because then we would have to allocate all buffers beforehand, but we want to allocate them on demand.

mara004:
https://github.com/pypdfium2-team/pypdfium2/pull/153

mara004:
Also linking #206

mara004:
Here's some pseudo code how the shm segment size could be calculated
```python
# scale: rendering scale
# bpp: bytes per pixel, e.g. 4 for BGRA with 1 byte per channel
# pdf: the document
# CACHE_MEM_LIMIT: maximum cache memory; may be overridden if a page is larger than that
# AVRG_CACHE_PAGES: the average number of pages the cache shall be able to store
page_mem_sizes = [w * h * scale**2 * bpp for (w, h) in (pdf.get_page_size(i) for i in range(len(pdf)))]
segment_size = max(min(average(page_mem_sizes)*AVRG_CACHE_PAGES, CACHE_MEM_LIMIT), max(page_mem_sizes))
```

PR #145, opened 2022-10-01
Add helpers for transform matrices and XObjects

mara004:
This enables callers to capture pages as XObjects, get a page object for the XObject, transform it, and add it to a page.
This could be used to implement a custom N-up compositor, for instance.


mara004:
Some things to consider:
* Should we add a `generate_content()` call to `insert_text()` for backwards compatibility, and remove it only with the next major release?
* Can we simplify the `PdfPageObject` initialisation code somehow?

mara004:
Found and reported https://crbug.com/pdfium/1905 while adding additional validation to `get_pos()`.

PR #148, opened 2022-10-05
Autoclose via weakref (and other improvements)

mara004:


mara004:
Let's merge this and continue in main.

PR #151, opened 2022-10-13
Support model for image objects

mara004:


mara004:
Note to self: want to add `pypdfium2 jpgtopdf` command-line interface to expose the functionality.

PR #152, opened 2022-10-14
Development branch for next API-breaking release

mara004:


mara004:
Logging a code comment here for loud thinking:

> Assumption: If the converted object uses the existing buffer instead of making a copy, it keeps the buffer object itself alive.
> It would be troublesome if it only protected the memory but discarded the buffer object, because then the finalizer of a pdfium-allocated bitmap buffer could be called prematurely.
> A possible counter-measure would be to transfer the finalizer from the buffer to the converted object, but we may only do that if the converter is guaranteed not to copy, because otherwise we would keep resources in memory unnecessarily.

mara004:
I think an approach to consider might be to never return the buffer directly, but always the info and store the buffer in an info attribute (and the finalizer in another). (Then in turn the buffer may not reference the info anymore to avoid a reference cycle, but it's nicer to have the attributes on direct level anyway.)
Then the caller would have a way to safely control the object's lifetime. But we could still consider transferring the finalizer to the converted object as suggested above.

mara004:
Also something to consider: The caller might want to work with the raw bitmap again after the converted object is not used anymore, so providing control over bitmap/finalizer would be a good thing indeed.

mara004:
I think, in addition to the above considerations, the converters should always return the `PdfBitmap` object and not just the `PdfBitmapInfo`. Then the caller of the single-page renderer has full control over the buffer, the finalizer, the `FPDF_BITMAP` handle etc. For the multi-page renderer, we can then just return the `PdfBitmapInfo` in `PdfDocument._process_page`, for the `PdfBitmap` itself can't be sent to the other process.

mara004:
Hmm, the new selective pageobject finalization doesn't work as expected. Current state broken. Need to investigate and possibly revert some changes.

mara004:
Might this have got something to do with the `__new__` method and us setting a finalizer in `__init__`?

mara004:
Hmm, page object removal does not behave the way I expected.
Apparently, it is not possible to re-insert removed objects. But that's a bit confusing, because then there would be no need to transfer ownership of the FPDF_PAGEOBJECT to the caller and it could be freed directly on removal.

mara004:
Found various mistakes in the textpage support model. Planning to work on this soon.

mara004:
Hmm, I think maybe/probably we should add the `close()` methods back?

Of course automatic finalization is great as a back-up, but I'm not sure how Python's garbage collector works, and suspect that the timing might not be optimal, considering that Python has no idea how large C objects are.
So it can still make sense to close stuff like a bitmap explicitly to make sure the memory is released immediately, rather than waiting until the next garbage collection cycle.

`close()` methods do expose some kind of a risk if the caller does not close in correct order, or attempts to access a closed object, but I think it still makes sense to offer a higher level of control over memory management.

src/pypdfium2/_helpers/textpage.py
```diff
@@ -201,45 +160,21 @@ def get_charbox(self, index, loose=False):
...
     
-    def get_rectboxes(self, index=0, count=0):
-        """
-        Get the bounding boxes of text rectangles in the requested scope.
-        
-        Yields:
-            Coordinates for left, bottom, right, and top (as :class:`float` values).
-        """
-        n_rects = self.count_rects(index, count)
-        for index in range(n_rects):
```
mara004, line 217: This was very wrong. We can't just do `range(n_rects)` here if `index` is greater than 0. We also need to take into account how many rects there are before `index`. The new function in the test code shows how it should have been:
```python
    if search_result is None:
        return
    c_index, c_count = search_result
    r_index = textpage.count_rects(0, c_index)
    r_count = textpage.count_rects(c_index, c_count)
    rects = [textpage.get_rect(i) for i in range(r_index, r_index+r_count)]
    return rects
```

src/pypdfium2/_helpers/textpage.py
```diff
@@ -130,29 +104,18 @@ def get_text_bounded(self, left=None, bottom=None, right=None, top=None, errors=
...
+            count (int): Character count to consider (defaults to -1 for all remaining).
         Returns:
-            int: The number of text rectangles on the page.
+            int: The number of text rectangles for the given character range.
         """
-        
-        if self.n_chars == 0:
-            return 0
-        if count == 0:
-            count = self.n_chars
```
mara004, line 152: Should have been `self.n_chars - index` here, but it's more convenient to let PDFium do this, anyway.
Also, the default for count should always have been `-1`, for `get_text_range()`, too.

mara004:
Hmm, namedtuple is not particularly flexible because we can't do post-init (assigning additional attributes based on the input).
That makes it rather unsuitable for `PdfBitmapInfo`. Maybe we should use `attrs`, but I'd rather want to avoid a mandatory runtime dependency.

mara004:
Plans for improving CI:
* Workflows to test setup, source build, and documentation build on Windows/Linux/macOS
* Add an integrated test step in the release workflow to try the built wheels on Windows/Linux/macOS with all supported Python versions before publishing them

That would be a great deal of work, though, so I might postpone it until after the release of v4.

mara004:
Work started: https://github.com/pypdfium2-team/pypdfium2/commit/039d3a3cef6857c2e543a0352ca9818406a8cceb

mara004:
The ambition of an integrated test step with different hosts turns out to be very difficult due to github restrictions.

The pipeline we want to achieve is `build -> test -> publish`. The problem is that `publish` needs to base on the exact state left by `build`, so we can't just make it three separate jobs. What we would need is to switch context to a different job and then return to the initial job. Unfortunately, it looks like github doesn't support that.

If we would use separate `build` and `publish` jobs, then wheel builds and repository updates would come from two different contexts, which potentially risks a mismatch.

mara004:
For reference, here's the patch for the separate build and publish attempt:
[0001-Attempt-to-add-inline-testing-to-release-WF.patch.txt](https://github.com/pypdfium2-team/pypdfium2/files/9946523/0001-Attempt-to-add-inline-testing-to-release-WF.patch.txt)
I experimented a lot with reusable workflows, `benc-uk/workflow-dispatch` and anything else I could think of but didn't manage it without that nasty separation.

mara004:
Two more ideas:
* Separate `build` and `test` jobs, `publish` as optional part of `build`. `test` optional, too. Run workflow twice, first with test enabled and publish disabled, then the reverse. Definitely not as nice as it could theoretically be, though.
* Alternatively (allowing some separation): Use temporary branch or patch anyway, write the planned tag to a status file, read the contents into a GH action job output, then add the tag in `publish`.

mara004:
I filed a discussion about this on GitHub Community: https://github.com/community/community/discussions/38443

mara004:
I think we'll use a patch or a temporary branch. I believe a patch is sufficient because `wheel` should not include any VCS information. Then we can postpone all VCS work to the publish step.

mara004:
to remember: consider switching from selective imports in `__init__` files to `__all__` lists in the relevant source forms.

mara004:
Hmm, this is taking a lot longer than expected. I think I'll do some remaining cleanup and then finally merge this into main. We can then suspend releases until it's ready, doing the test suite update in a separate PR.

mara004:
I think I'll close this PR for now an re-open a new one when it's actually ready.

PR #153, opened 2022-10-26
Inter-process callback: experimental attempt

mara004:
Just here to save the patch if we might resume working on this later.


mara004:
todo: try pipes instead of queues

mara004:
Linking #206 - maybe that was responsible for the breakage?

PR #156, opened 2022-11-07
Improve release workflow (and other things)

mara004:


mara004:
Hmm, windows build apparently fails because a patch does not apply correctly, which is very strange because I double-checked patches are in sync with pdfium-binaries and applied in the same order.

This is what the error message say, anyway:
```
Checking patch public/cpp/fpdf_deleters.h...
error: while searching for:
#ifndef PUBLIC_CPP_FPDF_DELETERS_H_?
#define PUBLIC_CPP_FPDF_DELETERS_H_?
?
#include "public/fpdf_annot.h"?
#include "public/fpdf_dataavail.h"?
#include "public/fpdf_edit.h"?
#include "public/fpdf_formfill.h"?
#include "public/fpdf_javascript.h"?
#include "public/fpdf_structtree.h"?
#include "public/fpdf_text.h"?
#include "public/fpdf_transformpage.h"?
#include "public/fpdfview.h"?
?
// Custom deleters for using FPDF_* types with std::unique_ptr<>.?
?

error: patch failed: public/cpp/fpdf_deleters.h:5
error: public/cpp/fpdf_deleters.h: patch does not apply
Checking patch public/cpp/fpdf_scopers.h...
error: while searching for:
#include <memory>?
#include <type_traits>?
?
#include "public/cpp/fpdf_deleters.h"?
?
// Versions of FPDF types that clean up the object at scope exit.?
?

error: patch failed: public/cpp/fpdf_scopers.h:8
error: public/cpp/fpdf_scopers.h: patch does not apply
Checking patch public/fpdfview.h...
error: while searching for:
// Dictionary value types.?
typedef int FPDF_OBJECT_TYPE;?
?
#if defined(COMPONENT_BUILD)?
// FPDF_EXPORT should be consistent with |export| in the pdfium_fuzzer?
// template in testing/fuzzers/BUILD.gn.?
#if defined(WIN32)?
#if defined(FPDF_IMPLEMENTATION)?
#define FPDF_EXPORT __declspec(dllexport)?

error: patch failed: public/fpdfview.h:176
error: public/fpdfview.h: patch does not apply
Checking patch BUILD.gn...
error: while searching for:
  ]?
}?
?
component("pdfium") {?
  libs = []?
  configs += [ ":pdfium_strict_config" ]?
  public_configs = [ ":pdfium_public_config" ]?

error: patch failed: BUILD.gn:160
error: BUILD.gn: patch does not apply
Checking patch core/fxge/win32/cgdi_printer_driver.cpp...
error: while searching for:
#include "core/fxcrt/fx_memory.h"?
#include "core/fxcrt/fx_system.h"?
#include "core/fxcrt/retain_ptr.h"?
#include "core/fxge/cfx_font.h"?
#include "core/fxge/cfx_windowsrenderdevice.h"?
#include "core/fxge/dib/cfx_dibextractor.h"?
```

mara004:
And another awkward error on a `macos-latest` runner with Python 3.11:
```
RuntimeError: No pre-built binaries available for platform 'macosx_10_9_universal2' with libc implementation 'None'. [...]
```

mara004:
I've browsed the source code for `sysconfig.get_platform()` a bit and realized it just does different things than I expected.
This explains the `10_9` part at least. The return of `universal2` is also understandable with the current implementation, but I don't quite get yet why only on Python 3.11.

Anyway, I think we need to move away from `sysconfig.get_platform()` if that's how it behaves. Probably use the `platform` module instead?

mara004:
While this implementation of the release action works, I think it might be cleaner to avoid a temporary branch and transfer a patch instead, but that would require splitting up the current autorelease script into two stages (stage 1: update version & changelog, create patch; stage 2: apply patch, generate release notes).

I will keep things as is for now because it's simpler, but that's an idea to keep in mind for the future.

mara004:
I don't 100% like the current state, but no doubt it adds additional safety to our release process, so I'll merge it in.
If we have an idea for a nicer structure, we can do it in following PRs.
Also keep in mind the windows sourcebuild still needs fixing, but I'm not quite sure about the source of the problem.

mara004:
As a reminder, I think we should also add a step to test the sdist to make sure all essential files are included.

mara004:
Windows sourcebuild has been fixed.

PR #174, opened 2023-02-14
readme (example comment): Fix order of tuple returned by searcher.get_next()

Lewistrick:
It incorrectly stated that it returns (left, right, bottom, top) tuples. Changed to (left, bottom, right, top).


mara004:
devel-new may need rebasing to include this commit

PR #220, opened 2023-05-14
API breaking changes (TOC improvements)

mara004:
TODO: update test suite


src/pypdfium2/_cli/toc.py
```diff
@@ -24,18 +24,17 @@ def attach(parser):
...
-    toc = pdf.get_toc(
-        max_depth = args.max_depth,
-    )
+    toc = pdf.get_toc(max_depth=args.max_depth)
     
-    for item in toc:
-        state = "*" if item.n_kids == 0 else "-" if item.is_closed else "+"
-        target = "?" if item.page_index is None else item.page_index+1
+    for bm in toc:
+        count, index, (view_mode, view_pos) = bm.get_count(), bm.get_index(), bm.get_view()
```
mara004: In the future, `count` and `index` could be inlined below by assignment expression (operator `:=`, requires python >= 3.8).

mara004:
Theoretically, it would be good practice to do this change with a major release.
However...
* compared to rendering, the TOC API seems to have relatively few callers, so I guess the overall impact of the change is low
* I don't really want to delay merging this until more API breaking changes have accumulated.

PR #237, opened 2023-07-17
POC patch to handle mmap as ctypes array

mara004:
Handling mmap as ctypes array (FPDF_LoadMemDocument64) might be better than as buffer (FPDF_LoadCustomDocument), to avoid unnecessary callbacks.

The way I understand mmap is that we can effectively treat it like regular memory without the file actually being fully in memory (i. e. the memory is a virtual view of a buffered stream, managed by the OS - access of the memory span maps to the stream).

But I'm not sure. The available information is rather scarce - maybe `.from_buffer()` just loads the mmap into memory as a whole?

### DIFF
--- a/src/pypdfium2/_helpers/document.py
+++ b/src/pypdfium2/_helpers/document.py
@@ -4,6 +4,7 @@
 __all__ = ("PdfDocument", "PdfFormEnv", "PdfXObject", "PdfBookmark", "PdfDest")
 
 import os
+import mmap
 import ctypes
 import logging
 import functools
@@ -72,6 +73,8 @@ class PdfDocument (pdfium_i.AutoCloseable):
         
         # CONSIDER adding a public method to inject into data_holder/data_closer
         
+        # NOTE due to mmap + weakref.finalize idiosyncrasies, we cannot hold a reference to the actual input on class level, otherwise we run into "cannot close; exported pointers exist" trouble
+        
         self._orig_input = input
         self._password = password
         self._autoclose = autoclose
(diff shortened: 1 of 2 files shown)


src/pypdfium2/_helpers/document.py
```diff
@@ -72,6 +73,8 @@ def __init__(self, input, password=None, autoclose=False, to_close=None):
         
         # CONSIDER adding a public method to inject into data_holder/data_closer
         
+        # NOTE due to mmap + weakref.finalize idiosyncrasies, we cannot hold a reference to the actual input on class level, otherwise we run into "cannot close; exported pointers exist" trouble
```
mara004, line 76: Theoretically we could if we were able to `del self._input` (or `self._input = None`) on before closing, but we cannot access `self` in the finalizer callback, so we cannot get rid of the reference.
weakref is awkward here: we cannot access the object that is being finalized, but apparently it still exists at that time.

mara004:
This seems to be pretty much the only way of handling mmap as ctypes array (I tried a lot of different patterns), so the patch is good for what it does, but I'm nevertheless not sure if it is acceptable for main.

The strictness of python regarding views held on an mmap is a real nuisance.
With all ctypes it's the caller's job to manage lifetime of C referenced objects, so an option to disable these checks would help a lot here.

mara004:
Alternatively we could try to leave this entirely to the caller, but that might complicate things a lot, at least where shared memory is concerned...

mara004:
I'll close this for now, but it's good for the record.

mara004:
Linking two comments of myself on a stackoverflow post:
https://stackoverflow.com/q/53339931/15547292

An idea I discuss there is to turn the input attribute into a mutable object (e.g. list `_wrapped_input`) and create a property accessor around it (`@property def _input(): return _wrapped_input[0]`), so we can get rid of the reference in the finalizer using `wrapped_input.clean()`.

src/pypdfium2/_helpers/document.py
```diff
@@ -80,15 +83,15 @@ def __init__(self, input, password=None, autoclose=False, to_close=None):
         if to_close:
             self._data_closer.extend(to_close)
         
-        self._input, to_close = _preprocess_input(self._orig_input)
+        input, to_close = _preprocess_input(self._orig_input)
```
mara004, line 86: See https://github.com/pypdfium2-team/pypdfium2/pull/237#issuecomment-1642899392 for an idea how to keep input accessible on object level.

mara004:
Or maybe we can pass `self.__dict__` to the finalizer?

PR #241, opened 2023-07-23
Major setup improvements

mara004:


setupsrc/pypdfium2_setup/packaging_base.py
```diff
@@ -11,28 +11,31 @@
...
-RepositoryURL        = "https://github.com/pypdfium2-team/pypdfium2"
-PDFium_URL           = "https://pdfium.googlesource.com/pdfium"
-DepotTools_URL       = "https://chromium.googlesource.com/chromium/tools/depot_tools.git"
-ReleaseRepo          = "https://github.com/bblanchon/pdfium-binaries"
-ReleaseURL           = ReleaseRepo + "/releases/download/chromium%2F"
+# NOTE if you rename BindingsFileName, also rename or delete the instance in the `bindings/` dir
+
+BinarySpec_EnvVar      = "PDFIUM_BINARY"
+BinarySpec_VersionSep  = ":"
+BinarySpec_V8Indicator = "-v8"
```
mara004: probably a bit over-dynamic, and long names too, but nvm

mara004:
fixme: install/emplace from sourcebuild seems to be broken (perhaps already prior to this commit?)

mara004:
todo: want to handle v8 as separate dirs so we can switch quickly between v8/non-v8 builds
not super relevant in practice, but the current handling is a stylistic mistake

mara004:
also want to create bindings only once for all platforms, but maybe not in this PR since it's a bigger change

Issue #177, opened 2023-02-22
v4 release blocker: PyPI direct dependency issue


mara004:
(Test)PyPI upload fails since we've pinned ctypesgen to a fork:
https://github.com/pypdfium2-team/pypdfium2/actions/runs/4246742423/jobs/7383996003
see also https://stackoverflow.com/questions/54887301/how-can-i-use-git-repos-as-dependencies-for-my-pypi-package

mara004:
Dumping TestPyPI output for the record:
```
25hINFO     Response from https://test.pypi.org/legacy/:                           
         400 Invalid value for requires_dist. Error: Can't have direct          
         dependency: "ctypesgen @                                               
         git+https://github.com/mara004/ctypesgen@pypdfium2 ; extra == 'setup'" 
INFO     <html>                                                                 
          <head>                                                                
           <title>[40](https://github.com/pypdfium2-team/pypdfium2/actions/runs/4246742423/jobs/7383996003#step:9:41)0 Invalid value for requires_dist. Error: Can't have direct 
         dependency: "ctypesgen @                                               
         git+https://github.com/mara004/ctypesgen@pypdfium2 ; extra ==          
         'setup'"</title>                                                       
          </head>                                                               
          <body>                                                                
           <h1>400 Invalid value for requires_dist. Error: Can't have direct    
         dependency: "ctypesgen @                                               
         git+https://github.com/mara004/ctypesgen@pypdfium2 ; extra ==          
         'setup'"</h1>                                                          
           The server could not comply with the request since it is either      
         malformed or otherwise incorrect.<br/><br/>                            
         Invalid value for requires_dist. Error: Can&#x27;t have direct         
         dependency: &quot;ctypesgen @                                          
         git+https://github.com/mara004/ctypesgen@pypdfium2 ; extra ==          
         &#x27;setup&#x27;&quot;                                                
                                                                                
                                                                                
          </body>                                                               
         </html>                                                                
ERROR    HTTPError: 400 Bad Request from https://test.pypi.org/legacy/          
         Invalid value for requires_dist. Error: Can't have direct dependency:  
         "ctypesgen @ git+https://github.com/mara004/ctypesgen@pypdfium2 ; extra
         == 'setup'"                                                            
Error: Process completed with exit code 1.
```

mara004:
[...]

PR #249, opened 2023-08-25
add dependency: pillow

milahu:
without pillow, the runtime fails with

```
$ pypdfium2
ModuleNotFoundError: No module named 'PIL'
```

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,6 +18,9 @@ name = "pypdfium2"
 description = "Python bindings to PDFium"
 readme = "README.md"
 requires-python = ">= 3.6"
+dependencies = [
+    "pillow",
+]
 dynamic = ["version"]
 keywords = ["pdf", "pdfium"]
 authors = [


mara004:
This is actually a bug, PIL is meant to be optional. I don't want to define it as hard dependency.
I believe the bug is limited to the CLI (`__main__.py` is not imported with the library by default).
`py3 -c "import pypdfium2"` should work without PIL.

I will fix the CLI problem shortly.
Note, when calling features that do depend on PIL, it will just fail with an attribute access error on None.
This is more or less deliberate. IMO it's clear enough. Marking all PIL spots with an explicit error message is probably overkill and prone to oversights/inconsistencies.

mara004:
Thank you for reporting this!

milahu:
lets add pillow as an [optional dependency](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies) to pyproject.toml

```toml
[project.optional-dependencies]
pillow = ["pillow"]
```

so it can be enabled with `pypdfium2[pillow]`

mara004:
Not sure if this is reasonable, then you can equally do `(pip install) pypdfium2 pillow`, that's basically no difference.
To my understanding, `[project.optional-dependencies]` is mostly used for groups of optional dependencies that form one feature or task, e.g. many people have groups like dev, docs, or test. But I haven't seen it for individual deps yet.

Also PIL is extremely common, almost everyone has it installed already.

mara004:
Oh, and apart from that, we're consistently using requirements files instead of `[project.optional-dependencies]` because of issue #177 (PyPI doesn't allow `git+` pinning in `[project]`). I forgot.

PR #251, opened 2023-08-27
v5 development (api breaking)

mara004:


mara004:
I think we should backport some important backwards-compatible fixes to main.

mara004:
Needs a major rebase after our work in main. It was not only backporting, some new fixes for issues that were discovered on the way need to be transferred back here.

mara004:
I guess I should split this up in setup and helpers changes so we can go ahead with conda packaging earlier.

mara004:
Closing this PR for said reason.

PR #252, opened 2023-09-02
Finally try to fix GH pages auto-rebuild (closes #166)

### LINKED ISSUE #166: Build workflow does not automatically trigger GitHub Pages rebuild

mara004:
Just try and use the `workflow-dispatch` action.

During testing, this now resulted in an awkward double trigger, with the first one automatically being cancelled.
The first was from the testing branch, the second from the main branch, so my guess/hypothesis would be that the a) `workflow_run:` trigger from main captured, however b) it captures only when `build_packages.yaml` is invoked manually and not through the bot trigger.

### DIFF
--- a/.github/workflows/build_packages.yaml
+++ b/.github/workflows/build_packages.yaml
@@ -211,6 +211,11 @@ jobs:
       
       - name: Publish to PyPI
         run: twine upload -u __token__ -p ${{ secrets.PYPI_TOKEN }} --verbose dist/*
+      
+      - name: Trigger GH Pages rebuild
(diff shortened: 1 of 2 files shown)


mara004:
Yes, it actually worked.

mara004:
Ah yes, https://stackoverflow.com/a/58393457/15547292 says
> If you need the push event to trigger other workflows, use a repo scoped [Personal Access Token](https://help.github.com/en/articles/creating-a-personal-access-token-for-the-command-line)

So that's indeed why the `workflow_run` trigger didn't capture.

PR #253, opened 2023-09-02
Backports from development branch

mara004:


mara004:
Good in principle, but I expect this will be a pain to rebase.

mara004:
Hmm, while running the test suite locally I caught a non-deterministic abort in a subprocess several times.
I can't reproduce it anymore after updating numpy, but this could also be simply due to changed time behavior.
However, I don't see what should be wrong with these changes.

mara004:
Let's assume this was down to numpy for the time being. If this reoccurs we can investigate again (e.g. bisect the commits and revert/analyze the problematic one).

PR #263, opened 2023-10-07
Improve version integration

mara004:


mara004:
Note: need to think out a fix for RTD, it doesn't check out enough info for git describe. :(

mara004:
^ [RTD build.jobs post_checkout](https://docs.readthedocs.io/en/stable/config-file/v2.html#build-jobs) sound like a possible solution to fetch the latest tag - and maybe also get out of the odd detached head state.

PR #267, opened 2023-10-21
conda_bundled patch (squashed, rebased)

mara004:
Just posting as PR for the record, to be closed immediately.


mara004:
Actually, I'm wondering if we couldn't just merge this into main after some polishing.
While it won't be actively used, it might be interesting to retain this as a second option.

PR #268, opened 2023-10-23
Conda packaging (WIP)

mara004:


conda/bundle/recipe/meta.yaml
```diff
@@ -0,0 +1,61 @@
...
+
+# NOTE
+# This is an attempt to bundle the pdfium binaries with conda packages, like we do for pypi wheels.
+# However, conda does not support CPU specific but Python version independent packages, meaning we'd have to build for each Python separately, so the better solution is probably to unbundle pdfium.
+# For now, let's retain this passively as a second option just in case a situation arises where bundling would be desired anyway.
+# - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+
+# FIXME since conda isolates each build, our data/ cache is no use...
+
+# usage: conda build conda/bundle/ --output-folder conda/bundle/out/
```
mara004, line 12: missing: VERSION=..

conda/helpers/recipe/meta.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+requirements:
+  build:
+    - git
+    - python
+    - pip
+    - setuptools
+    - wheel !=0.38.0,!=0.38.1
+  run:
+    # NOTE You can set a lower pdfium boundary should the necessity arise. Currently the limit for most features is probably so low that it's not significant. pdfium's API has been very stable in the past.
```
mara004, line 31: also note that pdfium-binaries is quite new on conda

mara004:
For the record, a note regarding the pdfium-binaries pin in pypdfium2_raw:
```
One could think about allowing lower versions if headers did not change.
While this should be ABI-safe, the problem is that it'd render the version file invalid,
and pdfium does not embed version into binaries, so it's hard to determine dynamically.
Also, the benefit would be quite low since we cannot programatically distinguish between
additions and changes to pdfium's public API. (And even if we could, it would require
hasattr-guards to support older versions that lack new symbols, which is not very nice.)
```

mara004:
I believe the packaging code is basically ready to be merged.
CI's not done yet but we could do that in a follow-up PR.

adam-huganir:
Doneso, under `adam@huganir.com`

mara004:
Thanks; unfortunately it looks like I can't add members by mail address. Would you please tell me the public user name instead? You might have to set that up under `Settings -> Public Profile` first.

adam-huganir:
> ^ ping @adam-huganir just in case you didn't get notified for the previous message no stress though, take your time :)

oops used the wrong anaconda site, I am `adam-huganir` on the .org one now

mara004:
Added, thanks!

PR #273, opened 2023-11-05
Add support for hermetic (internet-less) builds against binary pdfium releases

### LINKED ISSUE #272: Better instructions / support for offline building
### Description

I'm trying to package pypdfium2 for NixOS / [nixpkgs](https://github.com/NixOS/nixpkgs/) (the Linux distribution with currently by far the [most](https://repology.org/repositories/graphs) packages in).

Like many other distros, the build system has the requirement that builds are hermetic, without network access.

For the beginning I'm trying to use the pre-built `pdfium` binaries provided.

But the build system (paths down from `setup.py`) still tries to connect to the network in various places:

* downloading pdfium binaries
* fetching tags with git
* `urlretrieve()` in `build_pdfium_bindings()`; it has a `headers_dir=None` default argument, but no path in the code allows to set it to something not-`None`

I noticed the `prepared!` concept from the [README](https://github.com/pypdfium2-team/pypdfium2/tree/a397ff7efa71c2fbb2200548bd9be7f30de84d7c#setup-magic), but not sure that's what's needed, as that seems to skip `ctypesgen`.

**Would it be possible to provide setup.py with pre-downloaded files of whatever is needed to build a single pypdfium2 version?**

For example, from what I've read so far, for version `6097` it seems it would be sufficient to provide:

* pdfium binary: https://github.com/bblanchon/pdfium-binaries/releases/download/chromium%2F6097/pdfium-linux-x64.tgz
* headers: https://pdfium.googlesource.com/pdfium/+archive/refs/heads/chromium/6097/public.tar.gz

If one could just provide those to `setup.py` with flags or environment variables, it would solve the hermetic distro packaging issue.

nh2:
Fixes #272.

The env vars

```
PDFIUM_BINARIES_DIR
PDFIUM_HEADERS_DIR
```

are added which `setup.py` picks up.

See the commit messages for more details.

### DIFF
--- a/README.md
+++ b/README.md
@@ -583,7 +583,6 @@ As of early 2023, a single developer is author and rightsholder of the code base
 While using pypdfium2, you might encounter bugs or missing features.
 In this case, feel free to file an issue. If applicable, include details such as tracebacks, OS and CPU type, as well as the versions of pypdfium2 and used dependencies.
 
-Roadmap:
 * pypdfium2
   - [Issues panel](https://github.com/pypdfium2-team/pypdfium2/issues): Initial bug reports and feature requests. May need to be transferred to dependencies.
   - [Discussions page](https://github.com/pypdfium2-team/pypdfium2/discussions): General questions and suggestions.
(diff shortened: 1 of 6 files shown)


mara004:
Not acceptable, see https://github.com/pypdfium2-team/pypdfium2/issues/272#issuecomment-1793817300.

PR #275, opened 2023-11-07
Experiment with getting full version from pdfium-binaries record

mara004:
The design benefit of using the VERSION file is low (if any) IMO.

For refbindings and system, we still need the refs, so using the VERSION doesn't do much except add complexity.
It also looks cleaner to handle beforehand rather than in the loop, so I think we might be better off disk caching the remote refs to speed up consecutive process calls after all.

However, I like the style improvement of using a namedtuple (which is independent of the other change, though).


mara004:
I'll close this as not needed / bloat for now - the ls-remote performance is acceptable after all, so it doesn't merit the additional complexity.
We may want to pick the surrounding improvements later (namedtuple, external to_full?)

PR #278, opened 2023-11-22
Considerations regarding flags; try to add Skia

mara004:


mara004:
I'll pick something along these lines for main, but perhaps the considerations should go in an issue rather than in code comments...

setupsrc/pypdfium2_setup/packaging_base.py
```diff
@@ -394,8 +401,13 @@ def run_ctypesgen(target_dir, headers_dir, flags=[], guard_symbols=False, compil
     if not guard_symbols:
         args += ["--no-symbol-guards"]
     if flags:
-        args += ["-D"] + [f"PDF_ENABLE_{f}" for f in flags]
+        args += ["-D"] + [PdfiumFlagsDict[f] for f in flags]
     if Host.system == SysNames.windows:
+        # NOTE untested and probably not that useful (only if installing from source on windows)
+        # if we can get hold on the windows headers (or even just a stub providing the few types used by pdfium), we might be able to cross-compile with the --include option
```
mara004, line 407: ... not with --include, since we don't want to include all windows header symbols, just the ones used by pdfium.
So we'd need a way to pretend additional system includes to the preprocessor or something.

mara004: I figured out the `CPATH` env var can be used for this (or others), and was able to do a local cross-compilation POC using `CPATH` pointing to a dir with this fake header:
```c
// windows.h
typedef const struct HDC__ *HDC;
```

See also
https://stackoverflow.com/questions/558803/how-can-i-add-a-default-include-path-for-gcc-in-linux
https://stackoverflow.com/questions/34387096/getter-for-windows-handles#34387348

PR #280, opened 2023-11-22
conda_raw: handle rebuilds

mara004:


mara004:
Turns out it was quite unnecessary to use a separate file, see 9a1dcd358dd8114c60cc3da06c42b4fe5a8498ae for post-merge cleanup

PR #287, opened 2023-12-29
Presumably work around pdfium bug 2112

mara004:
I am however reluctant to make a release with this as the bug would still wreak breakage on downstreams that use a non-default bitmap maker. Which would be unusual, but legit API use.

Also the issue seems to be very circumstance-specific and the cause in the control flow is not yet known to me (although we seem to have narrowed down the commit and relation to stride), so I am not certain if this would truly work around all cases.


mara004:
CC https://github.com/pypdfium2-team/pypdfium2/commit/a1a1d378d1a5ced1c972fb6b29586bb718b11056

mara004:
Closing as the bug is now marked as fixed upstream.

PR #307, opened 2024-04-04
v5 devel branch

mara004:


src/pypdfium2/_helpers/page.py
```diff
@@ -262,7 +269,7 @@ def get_objects(self, filter=None, max_depth=2, form=None, level=0):
             :class:`.PdfObject`: A page object.
         """
```
mara004, line 275: note to folded code about: We may want to bump the default of max_depth from 2 to 15 to align with the extract-images CLI

src/pypdfium2/_helpers/page.py
```diff
@@ -217,16 +226,27 @@ def insert_obj(self, pageobj):
...
         if pageobj.page is not self:
-            raise ValueError("The page object you attempted to remove is not part of this page.")
+            raise ValueError("The pageobject you attempted to remove is not part of this page.")
+        
+        # https://pdfium-review.googlesource.com/c/pdfium/+/118914
+        if pageobj.type == pdfium_c.FPDF_PAGEOBJ_TEXT:
+            for wref in self._textpage_wrefs:
+                textpage = wref()
+                if textpage and textpage.raw:
+                    logger.warning(f"When removing a text pageobject, any textpage handles ought to be closed beforehand - auto-closing {textpage}.")
```
mara004: better: `f"Removing text pageobbject implicitly closes affected textpage {textpage}"`

semoal:
Changes looks quite good! Really impressed, is there any estimated ETA?

semoal:
> @semoal Thanks. There is no set time frame, and I'm currently immersed in some other things, but I'd be hoping to merge sooner than later, to avoid another stalled and diverged branch, as had unfortunately happened on the previous go at this.
> 
> However, this project has grown a bit over my head TBH, and I'm somewhat scared of breaking anything or making wrong API decisions, as this may affect many downstreams. Also, I'd like to address all API-breaking or otherwise significant changes I had in mind before going ahead with this.
> 
> Out of interest, is there any particular change you're looking forward to?

The flatten function exposed it's a pain-killer, we're struggling about to extract some information of a pdf with form fields filled.

mara004:
> The flatten function exposed it's a pain-killer, we're struggling about to extract some information of a pdf with form fields filled.

I see. FWIW, you can already use the semi-private `page._flatten()` if you make sure `pdf.init_forms()` was called on the parent pdf before page retrieval (ideally, directly after construction).
The bindings code is the same, just a check added and docs updated. You could also copy the `flatten()` implementation over into your own code.
Sorry for the inconvenience; this originated from a time where form initialization wasn't integrated properly.

src/pypdfium2/_helpers/pageobjects.py
```diff
@@ -278,7 +306,10 @@ def get_data(self, decode_simple=False):
         Returns:
             ctypes.Array: The data of the image stream (as :class:`~ctypes.c_ubyte` array).
         """
-        func = pdfium_c.FPDFImageObj_GetImageDataDecoded if decode_simple else pdfium_c.FPDFImageObj_GetImageDataRaw
+        func = {
+            False: pdfium_c.FPDFImageObj_GetImageDataRaw,
+            True: pdfium_c.FPDFImageObj_GetImageDataDecoded,
+        }[decode_simple]
```
mara004: TODO: want to change these `{False: ..., True: ...}[var]` dicts back to `... if var else ...` after all. so any object whose `bool()` is True can be used.
Also it seems like overkill to create a dict just for True/False.

hector-sherpas:
> > The flatten function exposed it's a pain-killer, we're struggling about to extract some information of a pdf with form fields filled.
> 
> I see. FWIW, you can already use the semi-private `page._flatten()` if you make sure `init_forms()` was called on the parent pdf before page retrieval (ideally, directly after construction). The bindings code is the same, just a check added and docs updated. You could also copy the `flatten()` implementation over into your own code. Sorry for the inconvenience; this originated from a time where form initialization wasn't integrated properly.

@mara004 I have tried using page._flatten() with the instructions you have given as follows: 
`pdf_doc = pdfium2.PdfDocument(pdf_path)
pdf_doc.init_forms()
for page_idx in page_range: # page_range --> 2
        page = pdf.get_page(page_idx)
        page._flatten(flag=pdfium_c.FLAT_NORMALDISPLAY) # return 1
        text_page = page.get_textpage()
        page = pdf.get_page(page_idx)
        page._flatten(flag=pdfium_c.FLAT_NORMALDISPLAY) # return 2
        text_page = page.get_textpage()
        ...
        total_chars = text_page.count_chars()
`
I have to repeat the _flatten() twice to get all the editable values ​​from the form.

number of characters without repeating _ flatten code --> total_chars # 4619
number of characters with repeating _ flatten code --> total_chars # 5014

mara004:
I can't really comment on that behavior as I'm only providing the bindings, and what the underlying APIs actually do is down to pdfium.

However, given that the second `_flatten()` call returns 2, which is equal to `pdfium_c.FLATTEN_NOTHINGTODO`, one would expect that it should be a no-op. Perhaps you just have to re-initialize the page handle? Or maybe call `page.gen_content()` after flattening?

hector-sherpas:
Finally, I just had to re-initialize the page handle. page.gen_content() option didn't work. Thanks

mara004:
That makes sense. I can add a note to the future docs that flattening invalidates existing handles to the page.

docs/devel/changelog_staging.md
```diff
@@ -4,7 +4,37 @@
...
+- `get_text_range()`: Removed implicit translation of default calls to `get_text_bounded()`, as pdfium reverted `FPDFText_GetText()` to UCS-2, which resolves the allocation concern. However, callers are encouraged to explicitly use `get_text_bounded()` for full Unicode support.
+- Removed legacy version flags.
+
+*Improvements and new features*
+- Added `PdfPosConv` helper and `PdfBitmap.get_posconv(page)` for bidirectional translation between page and bitmap coordinates.
+- Added `PdfObject.get_quad_points()` to get the corner points of an image or text object.
+- Exposed `PdfPage.flatten()` (previously semi-private `_flatten()`), after having found out how to correctly use it. Added check and updated docs accordingly.
+- Added context manager support to `PdfDocument`, so it can be used in a `with`-statement, because opening from a file path binds a file descriptor, which should be released explicitly, given OS limits on the number of open FDs.
+- If document loading failed, `err_code` is now assigned to the `PdfiumError` instance so callers may programatticaly handle the error subtype.
```
mara004: typo: `programatticaly` -> `programmatically`

src/pypdfium2/_helpers/pageobjects.py
mara004: done, I think

mara004:
fixme: i think today's changes introduced a regression with grayscale rendering; need to find and fix that

mara004:
Hmm, when trying to push, I get `fatal: unable to access 'https://github.com/pypdfium2-team/pypdfium2.git/': The requested URL returned error: 502`
Seems like github might have an outage? Or maybe I force-pushed too much and reached some limit?

mara004:
Note to self: Next time, we should really open a PR only when the branch is ready, and use
`https://github.com/pypdfium2-team/pypdfium2/compare/main...$BRANCH_NAME` until then.
Otherwise, the log gets quite polluted, esp. due to the force pushes.

PR #338, opened 2025-01-18
Enable ppc64le wheel

mgiessing:

## Description

This PR enables ppc64le manylinux wheels. However it requires to build and provide the `libpdfium.so` manually for the moment as the upstream will probably take a bit more time.

A _quick and dirty_ way is to replace the download variable in `update_pdfium.py` like this and point it to the uploaded ppc64le lib before building the wheel:

```bash
sed -i "s/fu = .*/fu = \"https:\/\/github.com\/user-attachments\/files\/18456063\/pdfium-linux-ppc64le.tgz\"/g" setupsrc/pypdfium2_setup/update_pdfium.py
```

Once pdfium-binaries provide a `pdfium-linux-ppc64le.tgz`, GH actions can be updated accordingly to automatically build manylinux wheels.

## Checklist

N/A

</details>
<details open><summary>Setup</summary>

- [x] This PR changes setup code (`setupsrc/`, `setup.py` etc.).
- [x] I have read through relevant doc sections [Installation -> From source][1] and [Setup Magic][2].
- [x] I have tested the change does not break internal callers.
- [x] I believe the change is maintainable and does not cause unreasonable complexity or code pollution.
- [x] The change does not try to shift a maintenance burden or upstream downstream tasks. *Keep handlers generic, avoid overly downstream-specific or (for us) effectively dead code passages.*
- [x] I have assessed that the targeted use case cannot reasonably be satisfied by existing means, such as [Caller-provided data files][3], or the change forms a notable improvement over possible alternatives.

[1]: https://github.com/pypdfium2-team/pypdfium2?tab=readme-ov-file#install-source
[2]: https://github.com/pypdfium2-team/pypdfium2?tab=readme-ov-file#setup-magic
[3]: https://github.com/pypdfium2-team/pypdfium2?tab=readme-ov-file#install-source-caller

</details>

N/A

</details>

### DIFF
--- a/setupsrc/pypdfium2_setup/packaging_base.py
+++ b/setupsrc/pypdfium2_setup/packaging_base.py
@@ -91,6 +91,7 @@ class PlatNames:
     linux_x86        = SysNames.linux   + "_x86"
     linux_arm64      = SysNames.linux   + "_arm64"
     linux_arm32      = SysNames.linux   + "_arm32"
+    linux_ppc64le    = SysNames.linux   + "_ppc64le"
     linux_musl_x64   = SysNames.linux   + "_musl_x64"
     linux_musl_x86   = SysNames.linux   + "_musl_x86"
     linux_musl_arm64 = SysNames.linux   + "_musl_arm64"
@@ -108,6 +109,7 @@ ReleaseNames = {
     PlatNames.linux_x86        : "linux-x86",
     PlatNames.linux_arm64      : "linux-arm64",
     PlatNames.linux_arm32      : "linux-arm",
+    PlatNames.linux_ppc64le    : "linux-ppc64le",
     PlatNames.linux_musl_x64   : "linux-musl-x64",
     PlatNames.linux_musl_x86   : "linux-musl-x86",
     PlatNames.linux_musl_arm64 : "linux-musl-arm64",
@@ -334,6 +336,8 @@ class _host_platform:
             return PlatNames.linux_arm64 if self._libc_name != "musl" else PlatNames.linux_musl_arm64
         elif self._is_plat("linux", "armv7l"):
             return PlatNames.linux_arm32
+        elif self._is_plat("linux", "ppc64le"):
+            return PlatNames.linux_ppc64le
         elif self._is_plat("windows", "amd64"):
             return PlatNames.windows_x64
         elif self._is_plat("windows", "arm64"):
@@ -365,6 +369,8 @@ def get_wheel_tag(pl_name):
         return _manylinux_tag("aarch64")
     elif pl_name == PlatNames.linux_arm32:
         return _manylinux_tag("armv7l")
+    elif pl_name == PlatNames.linux_ppc64le:
+        return _manylinux_tag("ppc64le")
     elif pl_name == PlatNames.linux_musl_x64:
         return "musllinux_1_1_x86_64"
     elif pl_name == PlatNames.linux_musl_x86:


mara004:
Thanks a lot for your efforts with ppc64le!

I'm not sure how to best handle this, as we're basically at the crossroads between v4 and an almost-ready v5, and I'm in doubt if merging this into v4 would make much sense at this stage.
Also, I'm a bit reluctant to merge this while we don't actually have pdfium-binaries for ppc64le *yet*.

However, I agree we'd need a better solution for packaging in the meantime, or more generally, packaging with a self-built binary for a platform other than those currently supported with pdfium-binaries.

If you are doing the packaging on ppc64le natively, I think may be able to use the following trick (without patches):
```bash
mkdir -p data/sourcebuild/
cp autorelease/bindings.py data/sourcebuild/bindings.py

# set $MY_BINARY to the path of the binary you built
cp $MY_BINARY data/sourcebuild/libpdfium.so

# see https://github.com/pypdfium2-team/pypdfium2?tab=readme-ov-file#install-source-caller for description of version fields
# set $MAJOR/$MINOR/$BUILD/$PATCH to the pdfium version
cat > "data/sourcebuild/version.json" <<END
{
  "major": $MAJOR,
  "minor": $MINOR,
  "build": $BUILD,
  "patch": $PATCH,
  "n_commits": 0,
  "hash": null,
  "origin": "sourcebuild",
  "flags": []
}
END

# to use the files from data/sourcebuild/ on setup
export PDFIUM_PLATFORM=sourcebuild

# to build a wheel
python3 -m build -wxn

# to re-tag the wheel from linux_$ARCH (as returned by sysconfig.get_platform()) to the usual manylinux format
export ARCH=ppc64le
python3 -m wheel tags --platform-tag manylinux_2_17_$ARCH.manylinux2014_$ARCH dist/pypdfium2-*-py3-none-linux_$ARCH.whl
```

Please let me know if that works. Thanks!

mgiessing:
Yes, you're right - I just tried to make life _a little bit easier_ by already integrating the wheel-building logic, but I agree it should be merged once binaries are provided in a sustainable way :)

I'll go forward with the process you mentioned in the meantime and try to upstream as much as possible to pdfium itself and pdfium-binaries project!

mara004:
Hey @mgiessing,
have you been able to look into upstreaming this to pdfium-binaries yet?
Concerning the sysroot and libclang-rt (https://github.com/bblanchon/pdfium-binaries/issues/187#issuecomment-2595252470), I thought maybe you could make a new workflow to create and upload these to the GH packages area, and use them in the patches?

mgiessing:
Hi, so the first part is done - the [sysroot PR](https://chromium-review.googlesource.com/c/chromium/src/+/6187330) is merged :)

However the libclang runtime depends on the [sysroot json](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/tools/clang/scripts/build.py#559) which is not yet updated in build script itself and I'm not sure if I can just update the sysroots without any approval to the newest version.

Once that is done I believe we are able to build the package for ppc64le

mara004:
Ah, I see, thanks for the info!
Yes, trying to upstream this to pdfium itself (rather than just adding the patches into pdfium-binaries) is probably cleaner, but it will take longer.

> However the libclang runtime depends on the [sysroot json](https://chromium.googlesource.com/chromium/src/+/refs/heads/main/tools/clang/scripts/build.py#559) which is not yet updated in build script itself and I'm not sure if I can just update the sysroots without any approval to the newest version.

Does this mean we basically have to wait for Google to update the sysroots?

mara004:
@mgiessing FYI, roughly 1y later, we now have first-class support for ppc64le, as of 5.3.0.
The glibc variant is built at pdfium-binaries using upstream's `ppc64` target and some minor patching.
The musl variant is built here using the native sourcebuild in a cibuildwheel/pypa container, via static cross-compiler.

PR #344, opened 2025-02-26
Replace `./run` script with a `justfile`

mara004:


mara004:
A possible alternative might be the [`task`](https://taskfile.dev/usage/) command runner.

Issue #346, opened 2025-03-09
ctypes.POINTER memory leak in `_buffer_reader` / `_buffer_writer`


richardhundt:
### Checklist

- [x] I confirm to be using an **official package** of `pypdfium2` from `PyPI` or `GitHub/pypdfium2-team`.

### Description

In the callbacks for reading and writing PDF files, `ctypes.POINTER` objects are being created at runtime:

```python
        block = ctypes.cast(data, ctypes.POINTER(ctypes.c_ubyte * size))
```

This is problematic since pointer types are cached, and since `size` can vary in the last chunk, there's an unbounded growth of the type cache.

This leads to a 500k process eventually growing to 2.6Gb after ~8hrs of read/write operations on PDFs.

Running `tracemalloc.compare_to` confirms that this is where all the memory is going. 

A workaround is to call `ctypes._reset_cache()` after closing the PDF (or after context `__exit__`), but that seems like a bit of a hack. 
I'm not really familiar with the CFFI, so not sure what the canonical way of reading/writing buffers is given a pointer and a size, but it seems like there should be another way which doesn't require declaring pointer types whenever you read a chunk into a buffer. No idea.

Either way, memory is stable if you do a `ctypes._reset_cache()` when releasing PDF resources after I/O operations.

### Install Info

```shell
Python 3.10.12 (main, Feb  4 2025, 14:57:36) [GCC 11.4.0]
Linux-6.8.0-1024-aws-x86_64-with-glibc2.35
pypdfium2 4.30.1
pdfium 133.0.6899.0 at /opt/app/code/_venv/lib/python3.10/site-packages/pypdfium2_raw/libpdfium.so

Name: pypdfium2
Version: 4.30.1
Summary: Python bindings to PDFium
Home-page: https://github.com/pypdfium2-team/pypdfium2
Author: pypdfium2-team
Author-email: geisserml@gmail.com
License: BSD-3-Clause, Apache-2.0, PdfiumThirdParty
Location: /opt/app/code/_venv/lib/python3.10/site-packages
Requires: 
Required-by:
```

mara004:
Well, we could do
```python
addr = ctypes.cast(p_buf_first, ctypes.c_void_p).value
buf = (ctypes.c_char * size).from_address(addr)
```
but this is a bit murky as it mixes up pointer and address. Pointer values are usually memory addresses, but they don't necessarily have to be according to the C spec. Yet, in practice, we could probably assume they are.

[...]

PR #347, opened 2025-03-09
Avoid creation of sized POINTER types at runtime

mara004:
CC https://github.com/pypdfium2-team/pypdfium2/issues/346


src/pypdfium2/internal/utils.py
```diff
@@ -45,9 +45,9 @@ def __init__(self, buffer):
         self.buffer = buffer
     
     def __call__(self, _, position, p_buf_first, size):
-        p_buf = ctypes.cast(p_buf_first, ctypes.POINTER(ctypes.c_char * size))
+        buf = (ctypes.c_char * size).from_address( ctypes.addressof(p_buf_first.contents) )
```
mara004: Fun fact: This is actually a regression introduced with [`6907ddf`](https://github.com/pypdfium2-team/pypdfium2/commit/6907ddf46ea21562c8b3a312be8f1e00b148b9f1#diff-8bf4cfeb75e07c78b612600d786fe4857093f107394f0f882d13184477ddca9eL35). Historically, we had done it with `from_address()` already.

(See [here](https://github.com/pypdfium2-team/pypdfium2/blame/cc0c2b619b0f918995d1b34aa4c91808de047262/src/pypdfium2/_helpers/_opener.py) and click "Blame prior to change" on the line in question.)

mara004: In terms of versions, it was right until `2.11.0 (2022-08-22)`, and regressed with `3.0.0b1 (2022-09-24)`. The regression has then been present until now (`5.0.0b1`).

mara004:
@richardhundt Could you test this patch? (You may need to update your code to the upcoming v5 API first.)

richardhundt:
> @richardhundt Could you test this patch? (You may need to update your code to the upcoming v5 API first.)

Sure, I'll test it tomorrow. Thanks!

PR #350, opened 2025-04-09
Lean build script

mara004:


mara004:
This is experimental; setup integration is yet TBD.

PR #352, opened 2025-04-17
test_sourcebuild: try to fix clang build

mara004:


mara004:
Finally all green: https://github.com/pypdfium2-team/pypdfium2/actions/runs/14506802708
This wasn't the main issue, but I should really have realized earlier to install `llvm ldd` instead of `llvm-18 ldd-18` and manually relinking. This was caused by me searching in the wrong release on packages.ubuntu.com and (wrongly) assuming these would refer to another version.

PR #353, opened 2025-04-18
setup: look for system pdfium if we don't have binaries

mara004:


mara004:
I think I'll merge this in; we can do license handling and target extraction in follow-up PRs.

PR #357, opened 2025-05-06
Include build-specific licenses aggregated by pdfium-binaries

mara004:


mara004:
TODO: Migrate all internal users of `LicenseRef-PdfiumThirdParty.txt` to the collected build licenses.

PR #360, opened 2025-06-02
Verify pdfium-binaries packages via SLSA

mara004:
CC https://github.com/bblanchon/pdfium-binaries/issues/206


mara004:
Fixme: this currently breaks downloading older versions that don't have the provenance file.

mara004:
Also, we need to install the slsa-verifier in the workflow.

Issue #328, opened 2024-09-30
Improve setup capabilities to avoid caller-provided data files in common use cases


mara004:
### Checklist

- [X] I confirm this is **not a question or feature request**. *Otherwise, use the Discussions page.*
- [X] I confirm this is **not an issue encountered with an installed build of pypdfium2, but about some other aspect of the project** (specify below). *Otherwise, use one of the package templates (PyPA/conda), even if you believe this is not a package-specific issue.*
- [X] I confirm this is **not about an unofficial build of pypdfium2**. *We do not support third-party builds, and they are not eligible for a bug report.*

### Reason for Generic issue (keyword/topic)

setup

### Description

Currently, the only way to do offline setup is using caller-provided data files, which is inconvenient, particularly with ctypesgen integration.

See the ideas in https://github.com/pypdfium2-team/ctypesgen/issues/17#issuecomment-2384179391 for new setup options that might allow for offline setup with higher-level interface and internally-managed ctypesgen:
>  [...] taking a full pdfium version (to avoid web calls to resolve it from the build cipher), a path to custom pdfium headers (e.g. `$PDFIUM_HEADERS`), and adding new `system-path:` and `bundle:` strategies to `$PDFIUM_PLATFORM`.

yslflow:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

mara004:
Warning: Do NOT click on the link in the previous (now hidden) comment. It is very probably malicious.
The user has been permanently blocked from this organization and I have filed a GitHub report for spam or abuse.

If you need an example for the dangers of clicking on unknown/malicious links, read e.g. this: https://en.wikipedia.org/wiki/Pegasus_(spyware)#Discovery

mara004:
Hmm, admittedly though, #363 doesn't really address the original issue yet.
It merely changes/improves the way how to plug in your own data files, but it doesn't actually lower the need for producing your own files just yet. However, exposing the `system-find` strategy that got recently added with #353 is a good first step in that direction.
Basically, we still need a way for the caller to supply the full version (thus bypassing a web call), and to set custom runtime libdirs, for systems such as NixOS that use special slots for the binaries.

mara004:
[...]

PR #363, opened 2025-06-25
setup: introduce subtargets

mara004:
Expose all setup targets through `emplace.py`, particularly the system pdfium search.
This also removes the `prepared!...` strategy in favor of `data/sourcebuild` and `data/system` passthrough.
Progresses #328.

*Update: Note, this PR now depend on https://github.com/pypdfium2-team/ctypesgen/pull/32, which needs to be merged in sync.*


mara004:
TBD: update readme and changelog to reflect the changes

mara004:
Somewhat unrelated, but I wonder, should we commit the ctypesgen pin to the repository on release as well, and then undo it after the release commit?
Seeing as one might as well install from the repository rather than from an sdist.

In general, there is a bit of a flaw in our ctypesgen integration: Logically, every pypdfium2 commit should be coupled to the then head of pypdfium2-ctypesgen. However, AFAIK, the only realistic way to achieve that would be integrating ctypesgen into pypdfium2 itself, which is not an option as it would mix up the histories.

mara004:
Okay, I guess I'm happy enough with this to merge it in.
If there's anything left, a follow-up PR will do.

PR #375, opened 2025-08-28
cibuildwheel mac/windows attempt

mara004:


mara004:
Crazy, this actually worked.
Windows binaries are rather big though, let's try if building with --single-lib might improve that.

mara004:
I think it might make sense to reinstall `--single-lib` as default after all, and change it to opt-out.
Seeing as building the Linux and macOS wheels with `--single-lib` should also result in some space savings.

mara004:
Final test run is here: https://github.com/pypdfium2-team/pypdfium2/actions/runs/17342112816/job/49237398547
I cancelled the emulated jobs after all native jobs had finished.

setupsrc/pypdfium2_setup/build_native.py
```diff
@@ -171,19 +171,20 @@ def get_sources(deps_info, short_ver, with_tests, compiler, clang_path, single_l
...
-                is_regex=False, exp_count=1,
-            )
-            autopatch(
-                PDFIUM_DIR/"public"/"fpdfview.h",
-                "#if defined(COMPONENT_BUILD)",
-                "#if 1  // defined(COMPONENT_BUILD)",
-                is_regex=False, exp_count=1,
-            )
+        # bundle dependencies (e.g. abseil) into the pdfium DLL
+        # TODO in the future, we might want to extract separate DLLs for the imaging libraries (e.g. libjpeg, libpng)
```
mara004, line 175: I just realized this comment is out of place for the native sourcebuild, given that use_system_libjpeg and so on are set to true. This needs to be moved to the toolchained sourcebuild instead.

mara004: Done in #377.

PR #390, opened 2025-11-08
Add extra licenses for cibuildwheel builds

mara004:


mara004:
We might be able to drop most of these in the future by converting all targets to vendored libraries.
If we can use a cross-compiler instead of an emulated compiler that should be theoretically possible for all cibuildwheel targets.

PR #391, opened 2025-11-08
Feature: `textpage` font helpers

mykolaskrynnyk:
## Description

This request adds helper functions to extract font-related information from `PdfTextPage`. A related feature request can be found in #358. Even though there are downstream packages that implement similar functionality, I think these features really belong here.

The implementation relies on two functions from PDFium's Experimental API, namely `FPDFText_GetFontInfo` and `FPDFText_GetFontWeight`, which, according to [this discussion](https://issues.chromium.org/issues/353746891), may be removed in the future in favour of another existing API. A few things to note here:

- If the experimental functions do get removed, the proposed implementation can be updated to use alternative functions in a backward compatible manner such that the end users of `pypdfium2` will not be affected.
- The functions proposed in the discussion are themselves marked as Experimental API, so using them does not offer a stability benefit either. In fact, `FPDFFont_GetFontName` mentioned there no longer exists and has been split into two other functions.
- Most importantly, these alternative functions from `FPDFFont_*` namespace require a font object that can be obtained from a text object, e.g.,

  ```python
  text_object = pdfium_c.FPDFText_GetTextObject(<PdfTextPage>, index)
  font = pdfium_c.FPDFTextObj_GetFont(text_object)
  ```
  
  Doing this in each helper function separately creates a significant overhead when one needs to get font size, family and weight together, resulting in ~25% slower execution. This overhead seems to be minimal if the font object is obtained just once and reused for further calls to get font information. However, this means that there would need to be a single function, e.g., `get_font_info` that returns an object containing font-related information. To me, the inclusion of such a function seems redundant but perhaps something you would want to consider in the future.

Let me know if you any thoughts on this.

## Checklist

- [x] Any new test files are under a suitable license and have been registered in `REUSE.toml`. (Not applicable)

</details>

</details>

- [ ] This PR changes other things, namely: ... 

</details>

### DIFF
--- a/src/pypdfium2/_helpers/textpage.py
+++ b/src/pypdfium2/_helpers/textpage.py
@@ -160,8 +160,57 @@ class PdfTextPage (pdfium_i.AutoCloseable):
         if n_rects == -1:
             raise PdfiumError("Failed to count rectangles.")
         return n_rects
-    
-    
+
+    def get_font_size(self, index=0):
+        """
+        Get the font size of a particular character.
+
+        Parameters:
+            index (int): Index of the character of interest.
+        Returns:
+            float: The font size of the particular character in points. Note
+                that PDFium outputs 1.0 when the font size is unknown.
+        """
+        font_size = pdfium_c.FPDFText_GetFontSize(self, index)
+        return font_size
+
+    def get_font_name(self, index=0):
+        """
+        Get the font name of a particular character.
+
+        Parameters:
+            index (int): Index of the character of interest.
+        Returns:
+            str | None : The font name of the particular character, or None
+                if the font name cannot be obtained.
+        """
+        flags = ctypes.c_int()
+        n_bytes = pdfium_c.FPDFText_GetFontInfo(self, index, None, 0, flags)
+        if n_bytes == 0:
+            # Gracefully handle errors since escape characters trigger these
+            return None
+        buffer = ctypes.create_string_buffer(n_bytes)
+        pdfium_c.FPDFText_GetFontInfo(self, index, buffer, n_bytes, flags)
+        # The NUL terminator is chopped off automatically
+        font_name = buffer.value.decode("utf-8")
+        return font_name
+
+    def get_font_weight(self, index=0):
+        """
+        Get the font weight of a particular character.
+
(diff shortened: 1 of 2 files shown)


mara004:
When we add a `PdfTextObj` (subclass of existing `PdfObject` machinery in `pageobjects.py`, similar to `PdfImage`), that might also be an opportunity to add the `PdfTextObj.extract()` that was asked for in #325.

src/pypdfium2/_helpers/textpage.py
```diff
@@ -160,8 +160,57 @@ def count_rects(self, index=0, count=-1):
         if n_rects == -1:
             raise PdfiumError("Failed to count rectangles.")
         return n_rects
-    
-    
+
+    def get_font_size(self, index=0):
```
mara004, line 164: Does default `index=0` really make sense here, or might it be preferable to have the caller set the index explicitly?

mara004:
> My main concern with this approach is an annoyingly inconsistent usage. 

FWIW the split seemed rather logical to me? Font size is specific to the particular bit of text, whereas name/family and weight are properties of the font itself, and stay the same across different textobjects that use this font.

If I can spare the time, I might attempt to implement this on my own as drafted in https://github.com/pypdfium2-team/pypdfium2/pull/391#issuecomment-3506650858 or the second code block of your https://github.com/pypdfium2-team/pypdfium2/pull/391#issuecomment-3506724399.

mykolaskrynnyk:
I have already started working on the alternative implementation. I'll make another PR tomorrow. Let's see if that works.

mara004:
For the record, another advantage of going the `FPDFText_GetTextObject()` route may be that a text object is often a group of chars, rather than just a single char, so this probably means less FFI calls.

Whereas what is proposed in this PR would have to be called char by char.

mykolaskrynnyk:
Fine. Closing in favour of #392 .

mara004:
@mykolaskrynnyk I've also integrated the font helpers into pypdfium2's CLI: https://github.com/pypdfium2-team/pypdfium2/commit/666e99f317a8aaf6504511e76455bece6c9447c7

PR #392, opened 2025-11-08
Add draft font helpers

mara004:
May supersede #391

Implements #325 and #358

### DIFF
--- a/tests/test_textpage.py
+++ b/tests/test_textpage.py
@@ -4,6 +4,7 @@
 import re
 import pytest
 import pypdfium2 as pdfium
+import pypdfium2.raw as pdfium_c
 from .conftest import TestFiles
 
 
@@ -153,26 +154,34 @@ def test_get_text_bounded_defaults_with_rotation():
     assert len(text) == 438
 
 
-def test_font_helpers():
+@pytest.mark.parametrize(
+    "index,character,text,font_size,base_name,family_name,weight",
+    [
+        (0, "L", "Lorem ipsum dolor sit amet,", 16.0, "Ubuntu", "Ubuntu", 400),
+        (5, " ", "Lorem ipsum dolor sit amet,", 16.0, "Ubuntu", "Ubuntu", 400),
+        (27, "\r", None, None, None, None, None),
+        (28, "\n", None, None, None, None, None),
+        (43, "i", "consectetur adipisici elit,", 16.0, "Ubuntu", "Ubuntu", 400),
+    ],
+)
+def test_font_helpers(
+    index, character, text, font_size, base_name, family_name, weight
+):
     pdf = pdfium.PdfDocument(TestFiles.text)
     page = pdf[0]
     textpage = page.get_textpage()
     n_chars = textpage.count_chars()
     print(n_chars)
-    
-    textobj = textpage.get_textobj(0)
-    text = textobj.extract()
-    font_size = textobj.get_font_size()
-    assert text == "Lorem ipsum dolor sit amet,"
-    assert font_size == 16
-    
-    fontobj = textobj.get_font()
-    base_name = fontobj.get_base_name()
-    family_name = fontobj.get_family_name()
-    weight = fontobj.get_weight()
-    assert base_name == "Ubuntu"
-    assert family_name == "Ubuntu"
-    assert weight == 400
-    
-    no_textobj = textpage.get_textobj(27)
-    assert no_textobj is None
+
+    assert chr(pdfium_c.FPDFText_GetUnicode(textpage.raw, index)) == character
+    textobj = textpage.get_textobj(index)
+    if text is None:
+        assert textobj is None
+    else:
+        assert textobj.extract() == text
+        assert textobj.get_font_size() == font_size
+
+        fontobj = textobj.get_font()
+        assert fontobj.get_base_name() == base_name
+        assert fontobj.get_family_name() == family_name
+        assert fontobj.get_weight() == weight


src/pypdfium2/_helpers/pageobjects.py
```diff
@@ -155,6 +157,84 @@ def transform(self, matrix):
...
+    """
+    """
+    
+    def __init__(self, *args, textpage=None, **kwargs):
+        if textpage is not None:
+            kwargs.update(page=textpage.page, pdf=textpage.page.pdf)
+        super().__init__(*args, **kwargs)
+        self.textpage = textpage
+    
+    def get_text(self):
```
mara004: originally proposed as `.extract()` in #325, could rename to that

src/pypdfium2/_helpers/pageobjects.py
```diff
@@ -155,6 +157,84 @@ def transform(self, matrix):
...
+            api = pdfium_c.FPDFFont_GetFamilyName
+        else:
+            raise ValueError(f"{which!r}")
+        
+        bufsize = api(self, None, 0)
+        buffer = ctypes.create_string_buffer(bufsize)
+        
+        out_bufsize = api(self, buffer, bufsize)
+        if out_bufsize == 0:
+            raise PdfiumError(f"Failed to get font {which} name.")
```
mara004, line 224: todo: check for this case after the first call, and return None (or an empty string?) instead of raising error

mara004: Hmm, given this is `FPDFFont_GetBaseFontName` and not `FPDFText_GetFontInfo`, I think it's correct to raise an exception at this point, after all.

src/pypdfium2/_helpers/pageobjects.py
```diff
@@ -155,6 +157,121 @@ def transform(self, matrix):
...
+            kwargs.update(page=textpage.page, pdf=textpage.page.pdf)
+        
+        super().__init__(*args, **kwargs)
+        self.textpage = textpage
+        
+        # If there is a textpage, the object should not be freed, but we attach a parent holder finalizer
+        if textpage is not None:
+            self._close_func = PdfTextObj._holder_impl
+            self._ex_args = (textpage, )
+            self._attach_finalizer()
```
mara004: Hmm, I don't think this would work with `PdfPage.remove_obj()`, in the `textpage is not None` case.
`remove_obj()` would need to activate a passive `FPDFPageObj_Destroy()` finalizer, but this doesn't work when we already have an active holder finalizer instead...

Also, `remove_obj()` says any `PdfTextPage` handles to the page ought to be closed before removing a text object. So perhaps it is really the page that needs to stay alive, rather than the textpage?
I need to re-read the API docs...

mara004: Huh, `FPDFText_GetTextObject()` merely says
> The caller does not own the returned object.

That doesn't make clear whether the object is owned by the textpage, or the page.

mara004: Easiest approach for now might be to acknowledge this in the docs:
> [!NOTE]
> Objects obtained from `PdfTextPage.get_textobj()` do not support `PdfPage.remove_obj()` at this time.

mara004: Or perhaps just abandon the object holder attempts for now and rely only on the python references, because `PdfObject`s currently don't hold their page in finalizers, either. This is probably a task for another PR in the future.

src/pypdfium2/_helpers/pageobjects.py
```diff
@@ -155,6 +157,103 @@ def transform(self, matrix):
...
+        
+        bufsize = api(self, None, 0)
+        if bufsize == 0:
+            raise PdfiumError(f"Failed to get font {which} name.")
+        
+        buffer = ctypes.create_string_buffer(bufsize)
+        out_bufsize = api(self, buffer, bufsize)
+        assert out_bufsize == bufsize
+        
+        return buffer.value.decode("utf-8")
```
mykolaskrynnyk: I think there is some merit to having two separate functions for base font and family names just like in the upstream.

src/pypdfium2/_helpers/pageobjects.py
```diff
@@ -155,6 +157,103 @@ def transform(self, matrix):
...
+    
+    def get_font_size(self):
+        """
+        Returns:
+            float: Font size used by the object's text, in PDF canvas units (typically 1/72in).
+        """
+        r_size = ctypes.c_float()
+        ok = pdfium_c.FPDFTextObj_GetFontSize(self, r_size)
+        if not ok:
+            raise PdfiumError("Failed to get font size.")
```
mykolaskrynnyk, line 206: Just to note here that this seems to return `1.0` when the font size cannot be determined. For that reason, I would maybe consider departing from the upstream here and return `None` if it's `1.0` instead.

src/pypdfium2/_helpers/pageobjects.py
```diff
@@ -155,6 +157,103 @@ def transform(self, matrix):
...
+            str: The objects's text content.
+        """
+        bufsize = pdfium_c.FPDFTextObj_GetText(self, self.textpage, None, 0)
+        if bufsize == 0:
+            raise PdfiumError("Failed to get text from textobject.")
+        
+        buffer = ctypes.create_string_buffer(bufsize)
+        buffer_ptr = ctypes.cast(buffer, ctypes.POINTER(pdfium_c.FPDF_WCHAR))
+        out_bufsize = pdfium_c.FPDFTextObj_GetText(self, self.textpage, buffer_ptr, bufsize)
+        assert bufsize == out_bufsize
```
mykolaskrynnyk: Since this is a mere sanity check, it seems redundant to have it in the public API. This applies to all `assert` statements in this PR.

src/pypdfium2/_helpers/pageobjects.py
mara004: May be so, but the usage of the two underlying pdfium APIs is effectively the same, and I thought this was easier than splitting in 3 functions (shared implementation, base name, family name).
FWIW there are already other helpers that do something similar, e.g. `PdfImage.load_jpeg()`.

src/pypdfium2/_helpers/pageobjects.py
mara004: Can you clarify when that happens? Do you have an example?
Wouldn't the API return False (causing the error to be raised) when the font size cannot be determined?
I doubt changing `1.0` to `None` is an option, if that value could also be intentional?

mykolaskrynnyk:
Here is another thing to be consider. The proposed implementation works fine for actual text objects but breaks in weird ways when escape characters are encountered. In `test_font_helpers` try index `27`, which corresponds to a carriage return:

```python
    def test_font_helpers():
        pdf = pdfium.PdfDocument(TestFiles.text)
        page = pdf[0]
        textpage = page.get_textpage()
    
        textobj = textpage.get_textobj(27)
>       text = textobj.extract()
               ^^^^^^^^^^^^^^^
E       AttributeError: 'PdfObject' object has no attribute 'extract
```

src/pypdfium2/_helpers/pageobjects.py
mara004: Sanity checks have already been useful in the past, and you can disable them by running python in optimized mode.
As long as there are no strong reasons against the check, I'd be +0 on keeping it.

mara004:
Ok, switched to None, I figured it makes more sense. Added test case for char 27.

src/pypdfium2/_helpers/pageobjects.py
mykolaskrynnyk: Try printing any webpage and save it as PDF. What is odd is that when I save a Word file as PDF (using Best for printing option), I also get `1.0` for font size. But perhaps this is a macOS issue only. I also tried (re)exporting a `./tests/resources/text.pdf` as PDF using Preview. The result is also stripped of font sizes.

I wouldn't know. if returning `1.0` is a deliberate choice. What I do know is that one could use the transformation matrix to approximate the font size.

src/pypdfium2/_helpers/pageobjects.py
mara004: Okay, nvm. I took your advice and split it, see the commit below.

src/pypdfium2/_helpers/pageobjects.py
mara004: I take it that means the error isn't raised, only `1.0` is returned, right?
In that case, I'd just keep it at that and maybe document the behavior...

src/pypdfium2/_helpers/textpage.py
```diff
@@ -235,13 +235,13 @@ def get_rect(self, index):
...
         Returns:
-            PdfTextObj: A handle to the textobject that includes the char at *index*.
+            PdfTextObj | None: A handle to the textobject that includes the char at *index*, or None if it could not be resolved (e.g. escape character).
         Tip:
             Textobjects can also be obtained through :meth:`.PdfPage.get_objects`.
         """
         raw_obj = pdfium_c.FPDFText_GetTextObject(self, index)
         if not raw_obj:
-            raise PdfiumError(f"Failed to get textobject for char at index {index}.")
+            return None
```
mykolaskrynnyk: Not sure about this one. I think this part isn't the issue. The text object _does_ exist for escape characters. It is when you call `extract` that the error occurs.

src/pypdfium2/_helpers/pageobjects.py
mykolaskrynnyk: I don't remember encountering any errors with this function. I believe it just falls back to `1.0`. Perhaps something to clarify/raise upstream? Elsewhere they return `-1` or `0`.

src/pypdfium2/_helpers/textpage.py
mykolaskrynnyk: I have read your other comment about the pointer above. I see why this is the case now. Never mind then.

mara004: Nope, raw_obj is a null pointer, please retry.
The wrapper object being created at all was a consequence of the incorrect check.

*Sorry, your comment hadn't shown up on my end yet*

src/pypdfium2/_helpers/pageobjects.py
mara004: I can't reproduce this. I print the wikipedia main page to PDF using chrome and took the first character, it says font size `13.33`. Could you upload one of your samples, perhaps the `./tests/resources/text.pdf` re-exported using macOS Preview?

mara004: The docs for `FPDFTextObj_GetFontSize()` only say "Returns TRUE on success." which sounds like the API never returns False at this time. Possibly, these macOS tools actually set `1.0` in the PDF, and PDFium merely passes through that value?

mykolaskrynnyk: It might have something to do with how the font size is set in a PDF. I believe there is more than one way this could be done. But then again, this is not really my purview. If you decide to dig deeper and find out what's going on, let me know.

On a related note, for the file added in #393 , the `weight` values are also `0`.

mara004: Thanks. As this is about the behavior of pdfium itself, I would perhaps let the matter rest, as I don't want this to become too much of a time sink.

src/pypdfium2/_helpers/pageobjects.py
mara004: Ok, nvm, I've removed these assertions after all.
My initial idea here was to confirm that the bytes had been actually written to the buffer, but I then realized that's probably not how it works, and the API always returns that size.
That might be a possible enhancement for pdfium's API design in the future, give the caller confirmation that the data has been written.

PR #393, opened 2025-11-09
Additional tests for font helpers

mykolaskrynnyk:
## Description

This request parametrises the font test to cover more cases and adds a new test PDF. Related to #392 .

## Checklist

- This PR changes helpers code.
- Any new test files are under a suitable license and have been registered in `REUSE.toml`.

</details>

- I have at least skimmed through [Installation -> With setup][1] and subsections.
- I believe the change is maintainable and does not cause unreasonable complexity or code fragmentation.
- The change does not shift a maintenance burden or upstream downstream tasks. *Keep handlers generic, avoid overly downstream-specific or (for us) effectively dead code passages.*
- I have assessed that the targeted use case cannot reasonably be satisfied by existing means, such as [Caller-provided data files][2], or the change forms a notable improvement over possible alternatives.

</details>

- [x] This PR changes other things, namely: adds a file exported via Preview for which `get_font_size` returns `1.0`.

</details>

### DIFF
--- a/tests/test_textpage.py
+++ b/tests/test_textpage.py
@@ -4,6 +4,7 @@
 import re
 import pytest
 import pypdfium2 as pdfium
+import pypdfium2.raw as pdfium_c
 from .conftest import TestFiles
 
 
@@ -172,7 +173,7 @@ def test_font_helpers(
     n_chars = textpage.count_chars()
     print(n_chars)
 
-    assert textpage.get_text_range(index, 1) == character
+    assert chr(pdfium_c.FPDFText_GetUnicode(textpage.raw, index)) == character
     textobj = textpage.get_textobj(index)
     if text is None:
         assert textobj is None


mara004:
Thank you. The parametrization looks fine.
Also thanks for the file, but I'd say as long as it isn't actually used by a test case it should not be committed to the repository (perhaps attach to a comment instead?). Or if it were used, it would need to be registered in `REUSE.toml` with an SPDX license identifier.

tests/test_textpage.py
```diff
@@ -153,26 +153,34 @@ def test_get_text_bounded_defaults_with_rotation():
...
-    family_name = fontobj.get_family_name()
-    weight = fontobj.get_weight()
-    assert base_name == "Ubuntu"
-    assert family_name == "Ubuntu"
-    assert weight == 400
-    
-    no_textobj = textpage.get_textobj(27)
-    assert no_textobj is None
+
+    assert textpage.get_text_range(index, 1) == character
```
mara004: I think there's also a raw API to get only a single character, perhaps that would be more appropriate here?

[`get_text_range()`](https://pypdfium2.readthedocs.io/en/stable/python_api.html#pypdfium2._helpers.textpage.PdfTextPage.get_text_range) is rather heavy and has some issues, e.g. `get_text_range(index, 1)` can theoretically return an empty string in rare cases (though not here). See the note box in docs, and the issues referenced in the sources.

mykolaskrynnyk: The closest raw API I found is `FPDFText_GetUnicode` that turns out to be much faster indeed.

PR #394, opened 2025-11-13
setupsrc/base.py: split off corebase, buildbase and version

mara004:
`base.py` has become increasingly hard to navigate, and convoluted.
Attempt to address that situation.


mara004:
> I'm guessing we should extract the non-split related changes and commit them to main first, after all...

Done, see https://github.com/pypdfium2-team/pypdfium2/commit/b4f2bc6ecf39c65d6b27782919e0d0ee12af5ab8

PR #395, opened 2025-11-15
main.yaml: add musllinux to test matrix (broken attempt)

mara004:


mara004:
I opened this PR just to record the attempt. But it's flawed, and I don't think it makes sense to pursue this any further, as anything `pip install` is broken.

Probably cleaner to add a separate job `test_musllinux` using either `container: quay.io/pypa/musllinux_1_2_${{ matrix.arch }}:latest`, or maybe the chroot-based `setup-alpine` action, and getting python dependencies from the system.

.github/workflows/main.yaml
```diff
@@ -102,39 +102,61 @@ jobs:
...
       
       - name: Set up Python
         uses: actions/setup-python@v6
+        if: ${{ !matrix.container }}
         with:
           python-version: ${{ matrix.py }}
       
+      - name: Container preparation
+        if: ${{ matrix.container }}
+        # there are also the PyPA-supplied python builds in /opt/python/* that ship with pip out of the box, but they are not in PATH by default and not sure how to add them
```
mara004, line 157: Ah, I've read this could be done by appending to `$GITHUB_PATH`, see
https://stackoverflow.com/questions/68211182/how-to-set-system-path-variable-in-github-action-workflow/73981698

PR #396, opened 2025-11-20
Add warning for missing system fonts during PDF rendering

scyyh11:

# Add warning for missing system fonts during PDF rendering

- Introduced a new method `_check_missing_fonts` to identify non-embedded fonts and log warnings if found.
- Added `is_embedded` method to `PdfFont` class to determine font embedding status.
- Enhanced logging to include text samples using non-embedded fonts for better context.

## Description

This PR adds a warning system to alert users when PDF pages contain non-embedded fonts that may not render correctly if the required system fonts are not available. When `PdfPage.render()` is called, the implementation now:

1. Scans the page for text objects using non-embedded fonts
2. Extracts sample text content that uses each problematic font
3. Logs a warning message listing the fonts and the text content that may be missing from the rendered image

The warning helps users identify when text might be silently omitted from rendered images due to missing system fonts, which was previously undetectable.

**Implementation details:**
- Uses PDFium's `FPDFFont_GetIsEmbedded()` API to check font embedding status
- Creates a temporary textpage to extract text content from text objects
- Limits text samples to 3 per font (with "and X more" indicator) to keep warnings readable
- Truncates very long text samples (>100 chars) for readability
- Gracefully handles edge cases (empty text, extraction failures, missing API)

**Example warning output:**
```
Page contains non-embedded fonts that may not render correctly if not available on the system: Arial (used in: "Hello", "World", "Test"); TimesNewRoman (used in: "Document", "Title" (and 2 more)). Text using these fonts may be missing from the rendered image.
```

## Checklist

- [ ] The change comes with sufficient test cases to confirm correct functionality.

</details>

- I have at least skimmed through [Installation -> With setup][1] and subsections.
- I believe the change is maintainable and does not cause unreasonable complexity or code fragmentation.
- The change does not shift a maintenance burden or upstream downstream tasks. *Keep handlers generic, avoid overly downstream-specific or (for us) effectively dead code passages.*
- I have assessed that the targeted use case cannot reasonably be satisfied by existing means, such as [Caller-provided data files][2], or the change forms a notable improvement over possible alternatives.

</details>

- This PR changes other things, namely: ... 

</details>

*mara004 edit: improve markdown formatting*


src/pypdfium2/_helpers/page.py
```diff
@@ -459,6 +459,32 @@ def render(
...
+                    warnings.append(f"{font_name} (used in: {text_display})")
+                else:
+                    # Fallback if we couldn't extract any text samples
+                    warnings.append(f"{font_name} (text content unavailable)")
+            
+            font_list = "; ".join(warnings)
+            logger.warning(
+                f"Page contains non-embedded fonts that may not render correctly if not available on the system: {font_list}. "
+                f"Text using these fonts may be missing from the rendered image."
+            )
```
mara004: Thank you for providing this code, I was able to take a quick look after all.

Could you factor out this implementation into a separate method to avoid lengthening the main `page.render()` method? Also, it may be better to make this optional. Opt-in, I think, as this potentially involves many API calls.

src/pypdfium2/_helpers/page.py
```diff
@@ -495,6 +521,64 @@ def _auto_bitmap_format(page, fill_color, grayscale, prefer_bgrx, maybe_alpha):
...
+                font = obj.get_font()
+                if not font.is_embedded():
+                    font_name = font.get_base_name()
+                    
+                    # Try to extract text content
+                    text_content = None
+                    try:
+                        # Create a PdfTextObj with textpage to extract text
+                        # We create a new wrapper with textpage since the original obj doesn't have one
+                        text_obj_with_textpage = PdfTextObj(obj.raw, textpage=textpage)
```
mara004: WDYT, should we maybe add a `textpage` passthrough option to `PdfPage.get_objects()` to avoid the need to re-instantiate the `PdfTextObj`s ?

src/pypdfium2/_helpers/pageobjects.py
```diff
@@ -251,6 +251,16 @@ def get_weight(self):
...
             raise PdfiumError("Failed to get font weight.")
         return weight
+    
+    def is_embedded(self):
+        """
+        Returns:
+            bool: True if the font is embedded in the PDF, False if it needs to be loaded from the system.
+        """
+        if not hasattr(pdfium_c, 'FPDFFont_GetIsEmbedded'):
+            # If the API is not available, assume embedded to avoid false warnings
```
mara004, line 261: Do you have a pointer in which pdfium version this API was added?

src/pypdfium2/_helpers/page.py
mara004: Or maybe just have it as a standalone public API, which the caller can invoke before `.render()` ?

src/pypdfium2/_helpers/page.py
```diff
@@ -495,6 +521,64 @@ def _auto_bitmap_format(page, fill_color, grayscale, prefer_bgrx, maybe_alpha):
...
+    Check for non-embedded fonts on the page that may not render correctly.
+    
+    Returns:
+        dict[str, list[str]]: Dictionary mapping font names (base names) to lists of text content that uses them.
+    """
+    missing_fonts = {}
+    textpage = None
+    try:
+        # Create a textpage to extract text content from text objects
+        textpage = page.get_textpage()
```
mara004: Maybe add passthrough here as well if a textpage is also needed on the caller side.

src/pypdfium2/_helpers/pageobjects.py
scyyh11: It was added in this commit:

* **PDFium commit:** `e617ccdd24622d9a026d03468d302ee7ea71d11e`
* **Title:** `Add new FPDFFont_GetIsEmbedded() API.`
* **Date:** 2022-05-11

src/pypdfium2/_helpers/page.py
scyyh11: For making this optional: my concern is that if we don’t run it by default, failures in the rendering step will remain completely silent — there’s currently no warning when something is missing. If the user doesn’t manually inspect the rendered result, they’ll never notice the failure. And even if they do notice, there’s no information about why the render failed.

src/pypdfium2/_helpers/page.py
```diff
@@ -444,6 +530,9 @@ def render(
         .. [#user_unit] Since PDF 1.6, pages may define an additional user unit factor. In this case, 1 canvas unit is equivalent to ``user_unit * (1/72)`` inches. PDFium does not currently provide an API to get the user unit, so this is not taken into account.
         """
         
+        # Check for non-embedded fonts and warn if found
+        self.warn_about_missing_fonts()
```
scyyh11: @mara004 For now, I’m still calling this method before rendering. But if you feel it’s unnecessary or would prefer it to be opt-in only, I’m happy to change it.

src/pypdfium2/_helpers/page.py
```diff
@@ -336,6 +340,88 @@ def flatten(self, flag=pdfium_c.FLAT_NORMALDISPLAY):
...
+            >>> missing = page.get_missing_fonts()
+            >>> if missing:
+            ...     print(f"Found {len(missing)} non-embedded fonts")
+            
+            >>> # Reuse an existing textpage
+            >>> textpage = page.get_textpage()
+            >>> missing = page.get_missing_fonts(textpage=textpage)
+            >>> textpage.close()
+        """
+        return _check_missing_fonts(self, textpage=textpage)
```
mara004: If calling `get_missing_fonts()` resolves to an identical `_check_missing_fonts()` call, the implementation can be moved here?

src/pypdfium2/_helpers/page.py
```diff
@@ -336,6 +340,127 @@ def flatten(self, flag=pdfium_c.FLAT_NORMALDISPLAY):
...
+            
+            font_list = "; ".join(warnings)
+            logger.warning(
+                f"Page contains non-embedded fonts that may not render correctly if not available on the system: {font_list}. "
+                f"Text using these fonts may be missing from the rendered image."
+            )
+        
+        return missing_fonts
+    
+
```
mara004: Maybe wrap in something like
```python
# SPDX-SnippetBegin
# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
# SPDX-SnippetCopyrightText: 2025 Yizhan Huang <your email here>

...

# SPDX-SnippetEnd
```

scyyh11: Should I also add it for the other file that I changed?

src/pypdfium2/_helpers/page.py
mara004: I agree the situation (silently incomplete output) isn't ideal, but I'd say not call it here, and instead add a note suggesting `warn_about_missing_fonts()` if it matters for your use case and you don't mind the increased API call surface and possible overhead.

IMHO the proper solution would be pdfium itself providing an API to register a "missing font" warner callback. Backtracking it from the caller side with the other public APIs, particularly through an FFI system, may not be ideal.

src/pypdfium2/_helpers/page.py
mara004: Thanks. You could add yourself as `SPDX-FileContributor` at the top of the other file, maybe.

scyyh11: Since the change in that file is just a small helper function, I’ll skip that for now.

src/pypdfium2/_helpers/page.py
scyyh11: That makes sense, thanks for the explanation.

src/pypdfium2/_helpers/page.py
```diff
@@ -336,6 +340,131 @@ def flatten(self, flag=pdfium_c.FLAT_NORMALDISPLAY):
...
+                    # Fallback if we couldn't extract any text samples
+                    warnings.append(f"{font_name} (text content unavailable)")
+            
+            font_list = "; ".join(warnings)
+            logger.warning(
+                f"Page contains non-embedded fonts that may not render correctly if not available on the system: {font_list}. "
+                f"Text using these fonts may be missing from the rendered image."
+            )
+        
+        return missing_fonts
```
mara004: I wonder, might it make more sense to move these two methods to `PdfTextPage`?

(Apologies for the chaotic review here, I'm still ill...)

src/pypdfium2/_helpers/page.py
```diff
@@ -336,6 +340,131 @@ def flatten(self, flag=pdfium_c.FLAT_NORMALDISPLAY):
...
+            >>> textpage.close()
+        """
+        missing_fonts = self.get_missing_fonts(textpage=textpage)
+        if missing_fonts:
+            warnings = []
+            for font_name in sorted(missing_fonts.keys()):
+                text_samples = missing_fonts[font_name]
+                if text_samples:
+                    # Limit number of text samples per font to keep warning readable
+                    max_samples = 3
```
mara004: nit: declare outside loop or maybe make parameter?

src/pypdfium2/_helpers/pageobjects.py
mara004: If we don't call this by default in `PdfPage.render()`, then we can assume the API exists and just let it fail otherwise.

mara004: FWIW I also tracked down the tag, `chromium/5059`. pypdfium2 already expects newer pdfium than that in any case.

src/pypdfium2/_helpers/page.py
scyyh11: I think it would only really make sense to move them to `PdfTextPage` if we expect these checks to be useful outside the `PdfPage` context. Right now I can't think of any concrete reuse case.

mara004: I believe a method that needs a textpage and has to do with text/fonts just logically fits better in `PdfTextPage` than `PdfPage`. Every textpage has a page, but not vice versa.
It avoids any "check for a passthrough textpage, else create an internal textpage that is thrown away" convolution from the roots.

src/pypdfium2/_helpers/page.py
mara004: For any other readers, see https://github.com/pypdfium2-team/pypdfium2/pull/396#discussion_r2564363034

src/pypdfium2/_helpers/page.py
```diff
@@ -336,6 +340,131 @@ def flatten(self, flag=pdfium_c.FLAT_NORMALDISPLAY):
...
+                            pass
+                        
+                        if font_name not in missing_fonts:
+                            missing_fonts[font_name] = []
+                        if text_content:
+                            missing_fonts[font_name].append(text_content)
+                except (PdfiumError, AttributeError):
+                    continue
+        except Exception:
+            pass
```
mara004: I might prefer to avoid broad exception catching here.
In particular, I'd say `Exception` and `AttributeError` are too generic and could theoretically mask e.g. a typo in the code.
Also, I think there is some redundancy in your current try/except hierarchy.

mara004: Even with `PdfiumError`, I'm not sure silently suppressing it is good.
Unless there's a concrete reason (like with `textpage.get_textobj()` and escape chars), it might fit the API design better not to do any exception catching here, and leave it up to the caller whether they want to guard for `PdfiumError`.

scyyh11:
> A small test case or something would be great, for coverage.
> 
> Also, this could be added to the `pypdfium2 render` CLI as an opt-out mechanism. (I can do that bit in a post-merge commit if you want.)

Sounds good, please add that to the render CLI in a follow-up commit.

src/pypdfium2/_helpers/page.py
```diff
@@ -336,6 +340,8 @@ def flatten(self, flag=pdfium_c.FLAT_NORMALDISPLAY):
         return rc
     
     
+    
+
```
mara004: nit: remove stale whitespace

mara004:
Thank you very much, I think I'll be able to merge this tomorrow.

src/pypdfium2/_helpers/page.py
```diff
@@ -277,6 +277,9 @@ def get_objects(self, filter=None, max_depth=15, form=None, level=0):
                 If None or empty, all objects will be provided, regardless of their type.
             max_depth (int):
                 Maximum recursion depth to consider when descending into Form XObjects.
+            textpage (PdfTextPage | None):
+                Optional textpage handle to attach to :class:`.PdfTextObj` instances.
+                This allows text extraction without re-instantiating text objects.
```
mara004, line 282: We should add some kind of check that the textpage passed in by the caller actually belongs to the page. `assert textpage.page is self` or something, I guess.

mara004: Though that check should only happen on first run, not in recursive invocations.

src/pypdfium2/_helpers/page.py
```diff
@@ -302,7 +305,7 @@ def get_objects(self, filter=None, max_depth=15, form=None, level=0):
                 raise PdfiumError("Failed to get pageobject.")
             
             # Don't register as child object, because the lifetime of pageobjects that are part of a page is managed by pdfium. The parent page should remain alive while a pageobject is used, but it seems unjustified to store countless of weakrefs just to lock pageobjects when the parent page is closed.
-            helper_obj = PdfObject(raw_obj, page=self, pdf=self.pdf, container=form, level=level)
+            helper_obj = PdfObject(raw_obj, page=self, pdf=self.pdf, container=form, level=level, textpage=textpage)
```
mara004, line 308: I wonder what's the best solution here. Perhaps add a `textpage` argument to `PdfObject` that is ignored but can be picked up by subclass?
An alternative might be checking `helper_obj.type == pdfium_c.FPDF_PAGEOBJ_TEXT` and post-assigning the textpage in that case, but I'm -1 on pulling in type if-checks here.

src/pypdfium2/_helpers/textpage.py
```diff
@@ -283,6 +283,95 @@ def search(self, text, index=0, match_case=False, match_whole_word=False, consec
...
+                        text_content = extracted[:97] + "..."
+                    else:
+                        text_content = extracted
+                
+                if font_name not in missing_fonts:
+                    missing_fonts[font_name] = []
+                if text_content:
+                    missing_fonts[font_name].append(text_content)
+        
+        return missing_fonts
```
mara004, line 328: Taking a closer look at what this function is doing, it just collects all non-embedded fonts – but even if the font is installed, this would currently raise an (erroneous) warning, right?
Is there any chance of checking whether the font is actually installed or not?
I'm not experienced with that part of the API, but I seem to remember something along `fpdf_sysfontinfo.h`.

mara004: Also, I'm not convinced unlimited sample collection / extraction is good.
Maybe move the max_samples mechanism here and make it a parameter? Caller can set `math.inf` if they really want unlimited samples.

src/pypdfium2/_helpers/pageobjects.py
```diff
@@ -251,6 +251,16 @@ def get_weight(self):
         if weight == -1:
             raise PdfiumError("Failed to get font weight.")
         return weight
+    
+    def get_is_embedded(self):
```
mara004: I might prefer just `is_embedded()`

src/pypdfium2/_helpers/page.py
scyyh11: How about using `**kwargs` ?

src/pypdfium2/_helpers/textpage.py
scyyh11: I'm afraid that we are not able to get the actual glyph or font file that the system used to render the text using the current API.

src/pypdfium2/_helpers/page.py
mara004: Would also need an if-check I think?

src/pypdfium2/_helpers/textpage.py
mara004: Oh, right, I didn't think about substitution...
Then maybe rename `get_missing_fonts()` to `get_nonembedded_fonts()` ?
Because a non-embedded font isn't necessarily misssing.

mara004: > I'm afraid that we are not able to get the actual glyph or font file that the system used to render the text using the current API.

Though, checking whether the non-embedded font is installed on the system might still be possible? Even if we can't find out about substitution, it would avoid erroneous warnings if the exact font is installed.

scyyh11: Sorry for the late response, I was having my finals last week. 

Yes, it is possible to check whether the non-embedded font is installed. However, from what I understand, pypdfium2 just delegates font lookup to PDFium, and PDFium in turn hands the request to the platform (fontconfig on Linux, GDI on Windows, CoreText on macOS, etc.) and uses whatever comes back. To check system fonts ourselves we’d have to:

- call into platform-specific APIs or tools (fontconfig on Linux, different APIs on Windows/macOS)

- somehow normalize the PDF font name to what those APIs expose (family vs full name vs PostScript name, localized names, style variants, etc.).

Even then, we still wouldn’t be guaranteed to be checking the same thing pypdfium2 uses internally, so the result could still disagree with what actually happens at render time. I’m a bit worried that this would add a fair amount of platform-specific complexity.

mara004: I was hoping pdfium might be exposing APIs for that job, but if it doesn't, I agree it's not feasible.

Lei suggested `FPDF_SYSFONTINFO` in https://crbug.com/464315000#comment4 but I'm not quite sure how it works or what exactly it does.

mara004:
Again I am not 100% sure if you used AI here or not, but since your other two PRs had AI involvement I will mark this as suspected AI too.

To state it again for any readers, AI is now banned from this project. See https://github.com/pypdfium2-team/pypdfium2/pull/413#issuecomment-4759951913

PR #403, opened 2026-02-16
base: use dict for platform -> wheeltag mapping

mara004:


mara004:
I'm rather +1 on this change.

PR #413, opened 2026-03-26
Add PdfFont.is_embedded() and PdfTextPage.get_font_substitutions()

scyyh11:
## Description

Add two new APIs for detecting non-embedded font substitutions, using PDFium-native APIs only (no platform-specific code):

### `PdfFont.is_embedded()`
Wraps `FPDFFont_GetIsEmbedded()`. Returns `True` if the font is embedded in the PDF, `False` otherwise. Raises `PdfiumError` on the `-1` failure case (rather than masking it as `True` via `bool()`).

### `PdfTextPage.get_font_substitutions()`
Iterates text objects on the page via `page.get_objects(filter=[FPDF_PAGEOBJ_TEXT])`. For each non-embedded font, compares:
- `FPDFFont_GetBaseFontName()` — what the PDF requested
- `FPDFFont_GetFamilyName()` — what PDFium actually resolved to

Returns `dict[str, set[str]]` — maps each non-embedded base font name to the set of family names PDFium resolved it to. Uses `set` because the same base name could theoretically resolve to different families across objects with different FontDescriptor flags. Returns empty dict if all fonts are embedded.

No `textpage` assignment is needed — `get_font()`, `is_embedded()`, `get_base_name()`, and `get_family_name()` all work without a textpage. The method is opt-in (not called during `render()`).

### Example

```python
pdf = pdfium.PdfDocument("document.pdf")
page = pdf[0]
tp = page.get_textpage()
subs = tp.get_font_substitutions()
# On Linux without CJK fonts:
# {'NotoSansCJKsc-Regular': {'Chrome Sans MM'}}
# On Linux with fonts-noto-cjk installed:
# {'NotoSansCJKsc-Regular': {'Noto Sans CJK SC'}}
# On a PDF with all fonts embedded:
# {}
```

The user inspects the mapping and judges — `Chrome Sans MM` means PDFium fell back to its generic substitute (font missing), while a real family name like `Arial` or `Noto Sans CJK SC` means the system resolved it correctly.

### Empirically verified substitution behavior

| Scenario | Base Name | Family Name |
|---|---|---|
| Embedded | Ubuntu | Ubuntu |
| Resolved from system | Helvetica | Arial |
| Resolved from system | STSong | Songti SC |
| PDFium built-in | Symbol | Chrom Symbol OTF |
| **Missing font** | **NotoSansCJKsc-Regular** | **Chrome Sans MM** |
| **Missing font** | **SimSun** | **Chrome Sans MM** |

### Tests
- `test_font_is_embedded` — embedded Ubuntu font returns `True`
- `test_font_is_embedded_failure` — null `FPDF_FONT` handle raises `PdfiumError`
- `test_get_font_substitutions_embedded` — text.pdf (all fonts embedded) → `{}`
- `test_get_font_substitutions_empty_page` — empty.pdf → `{}`
- `test_get_font_substitutions_fake_font` — handcrafted inline PDF with non-existent `FakeTestFont` → detected in substitutions

Verified on macOS and Linux (Docker). Also tested with the original bug file where `NotoSansCJKsc-Regular` → `Chrome Sans MM` on a bare Linux container.

## Checklist

</details>

</details>

- [ ] This PR changes other things, namely: ...

</details>

### DIFF
--- a/src/pypdfium2/_font_tracking.py
+++ b/src/pypdfium2/_font_tracking.py
@@ -1,0 +1,159 @@
+# SPDX-FileCopyrightText: 2025 geisserml <geisserml@gmail.com>
+# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+__all__ = ("get_font_requests", "clear_font_requests")
+
+import ctypes
+import threading
+import logging
+import pypdfium2.raw as pdfium_c
+import pypdfium2.internal as pdfium_i
+
+logger = logging.getLogger(__name__)
+
+# Module-level state — kept alive for the library lifetime.
+_default_info = None   # POINTER(FPDF_SYSFONTINFO) from FPDF_GetDefaultSystemFontInfo()
+_wrapper = None        # Our FPDF_SYSFONTINFO struct
+_cb_refs = []          # Strong refs to CFUNCTYPE objects to prevent GC
+_font_requests = {}    # {font_name_str: bool_found}
+_lock = threading.RLock()
+
+
+def _install_font_tracking():
+    """
+    Install a wrapper around PDFium's default system font info to intercept
+    MapFont/GetFont calls and log which fonts were found or missed.
+
+    Must be called after FPDF_InitLibraryWithConfig(). All fallible work
+    (getting default info, creating callbacks, building the wrapper) completes
+    before FPDF_SetSystemFontInfo() — the point of no return.
+    """
+    global _default_info, _wrapper, _cb_refs, _font_requests
+
+    # Step 1: Get a fresh default font info
+    default_info = pdfium_c.FPDF_GetDefaultSystemFontInfo()
+    if not default_info:
+        logger.debug("FPDF_GetDefaultSystemFontInfo() returned NULL; font tracking unavailable")
+        return
+
+    default = default_info.contents
+
+    # Step 2: Create all callback closures
+    # MapFont and GetFont intercept and log; all others are pure delegation.
+    # Intercepting callbacks use fail-open: delegate first, then try to log.
+    # An exception in logging must never affect the delegation result.
+
+    def _map_font(pThis, weight, italic, charset, pf, face, exact):
+        result = default.MapFont(default_info, weight, italic, charset, pf, face, exact)
+        try:
+            if face:
+                name = ctypes.string_at(face).decode("utf-8", errors="replace")
+                with _lock:
+                    if not _font_requests.get(name, False):
+                        _font_requests[name] = bool(result)
+        except Exception:
+            pass
+        return result
+
+    def _get_font(pThis, face):
+        result = default.GetFont(default_info, face)
+        try:
+            if face:
+                name = ctypes.string_at(face).decode("utf-8", errors="replace")
+                with _lock:
+                    if not _font_requests.get(name, False):
+                        _font_requests[name] = bool(result)
+        except Exception:
+            pass
+        return result
+
+    def _enum_fonts(pThis, mapper):
+        default.EnumFonts(default_info, mapper)
+
+    def _release(pThis):
+        if default.Release:
+            default.Release(default_info)
+
+    def _get_font_data(pThis, font, table, buffer, buf_size):
+        return default.GetFontData(default_info, font, table, buffer, buf_size)
+
+    def _get_face_name(pThis, font, buffer, buf_size):
+        return default.GetFaceName(default_info, font, buffer, buf_size)
+
+    def _get_font_charset(pThis, font):
+        return default.GetFontCharset(default_info, font)
+
+    def _delete_font(pThis, font):
+        default.DeleteFont(default_info, font)
+
+    # Step 3: Build wrapper struct and assign callbacks
+    wrapper = pdfium_c.FPDF_SYSFONTINFO()
+    wrapper.version = default.version
+
+    cb_refs = []
+    for fname, pyfunc in [
+        ("Release", _release),
+        ("EnumFonts", _enum_fonts),
+        ("MapFont", _map_font),
+        ("GetFont", _get_font),
+        ("GetFontData", _get_font_data),
+        ("GetFaceName", _get_face_name),
+        ("GetFontCharset", _get_font_charset),
+        ("DeleteFont", _delete_font),
+    ]:
+        pdfium_i.set_callback(wrapper, fname, pyfunc)
+        # Keep a strong ref to the ctypes callback object to prevent GC
+        cb_refs.append(getattr(wrapper, fname))
+
+    # Step 4: Assign module-level strong refs (before point of no return)
+    _default_info = default_info
+    _wrapper = wrapper
+    _cb_refs = cb_refs
+    _font_requests.clear()
+
+    # Step 5: Point of no return — install the wrapper
+    pdfium_c.FPDF_SetSystemFontInfo(_wrapper)
+
+
+def _reset_font_tracking():
+    """
+    Clear the font request log after FPDF_DestroyLibrary().
+
+    The wrapper, default_info, and callback refs are NOT cleared here —
+    they must remain alive during FPDF_DestroyLibrary() since PDFium calls
+    Release during shutdown. They become inert after destroy and are left
+    for process exit / GC.
+    """
+    with _lock:
+        _font_requests.clear()
+
+
+def get_font_requests():
+    """
+    Return the accumulated font resolution log.
+
+    Each entry maps a font name (as requested by PDFium via MapFont/GetFont)
+    to a boolean indicating whether the font was found on the system.
+
+    Font requests are accumulated across all page loads since library init
+    (or since the last call to :func:`clear_font_requests`).
+
+    Note:
+        The font names logged here are those passed to PDFium's MapFont/GetFont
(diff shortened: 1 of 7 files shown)


mara004:
Thank you for looking into this again.

Unfortunately I'm still -1 on taking this into pypdfium2. FWIW I think that `get_font_substitutions()` is improperly named, as what it actually does is just map base font name to family name(s). While the results may allow for some user-level empirical deduction, that sounds like an imprecise business and not a true solution to the problem at hand, finding out exactly which fonts are missing and how they were substituted, in a way that does not depend on human guesswork.

The implementation of `get_font_substitutions()` does not directly use C APIs and is relatively simple so it could just as well live on the caller side.

mara004:
Nevermind, `FPDF_SYSFONTINFO` doesn't work the way I hoped it might :/
Any attempt to call into a default handler from within our wrapping implementation promptly causes a crash.
I'm concluding that that's not how FPDF_GetDefaultSystemFontInfo() is meant to be used after all.

src/pypdfium2/_font_tracking.py
```diff
@@ -0,0 +1,159 @@
...
+    # Step 2: Create all callback closures
+    # MapFont and GetFont intercept and log; all others are pure delegation.
+    # Intercepting callbacks use fail-open: delegate first, then try to log.
+    # An exception in logging must never affect the delegation result.
+
+    def _map_font(pThis, weight, italic, charset, pf, face, exact):
+        result = default.MapFont(default_info, weight, italic, charset, pf, face, exact)
+        try:
+            if face:
+                name = ctypes.string_at(face).decode("utf-8", errors="replace")
```
mara004, line 50: dangerous: if face is a null pointer, `ctypes.string_at()` will crash.
IIRC, `ctypes.cast(..., ctypes.c_char_p).value` or something handles null pointers gracefully.

mara004: ah, nevermind, I see you're checking `if face:` above! fine then.

src/pypdfium2/_font_tracking.py
```diff
@@ -0,0 +1,159 @@
...
+    (getting default info, creating callbacks, building the wrapper) completes
+    before FPDF_SetSystemFontInfo() — the point of no return.
+    """
+    global _default_info, _wrapper, _cb_refs, _font_requests
+
+    # Step 1: Get a fresh default font info
+    default_info = pdfium_c.FPDF_GetDefaultSystemFontInfo()
+    if not default_info:
+        logger.debug("FPDF_GetDefaultSystemFontInfo() returned NULL; font tracking unavailable")
+        return
```
mara004, line 37: I wonder, do all our target platforms (linux, mac, windows, android, and a few others) have a default sysfont info? If not, we have a problem...

scyyh11: Seems android is the problem, on Android font resolution goes through CFX_AndroidFontInfo / CFPF_SkiaFontMgr internally, bypassing the FPDF_SYSFONTINFO callbacks entirely. I agreed with opt-in class design.

mara004: I took a quick look at upstream's code. It seems that `FPDF_GetDefaultSystemFontInfo()` indeed covers all platforms where pdfium internally supports system font mapping (makes sense).
So if `FPDF_GetDefaultSystemFontInfo()` returned NULL pdfium could not work with system fonts on that platform anyway.

*edit: added missing 'not'*

```cpp
FPDF_EXPORT FPDF_SYSFONTINFO* FPDF_CALLCONV FPDF_GetDefaultSystemFontInfo() {
  std::unique_ptr<SystemFontInfoIface> font_info =
      CFX_GEModule::Get()->GetPlatform()->CreateDefaultSystemFontInfo();
  if (!font_info) {
    return nullptr;
  }
  ...
```

```
$ grep -r CreateDefaultSystemFontInfo
fxge/android/fx_android_impl.cpp:  std::unique_ptr<SystemFontInfoIface> CreateDefaultSystemFontInfo() override {
fxge/apple/fx_apple_platform.cpp:CApplePlatform::CreateDefaultSystemFontInfo() {
fxge/apple/fx_apple_platform.h:  std::unique_ptr<SystemFontInfoIface> CreateDefaultSystemFontInfo() override;
fxge/linux/fx_linux_impl.cpp:  std::unique_ptr<SystemFontInfoIface> CreateDefaultSystemFontInfo() override {
fxge/win32/cwin32_platform.cpp:CWin32Platform::CreateDefaultSystemFontInfo() {
fxge/win32/cwin32_platform.h:  std::unique_ptr<SystemFontInfoIface> CreateDefaultSystemFontInfo() override;
fxge/cfx_gemodule.cpp:      g_pGEModule->platform_->CreateDefaultSystemFontInfo());
fxge/cfx_gemodule.h:    CreateDefaultSystemFontInfo() = 0;
fxge/fx_font_unittest.cpp:        CFX_GEModule::Get()->GetPlatform()->CreateDefaultSystemFontInfo());
```

mara004: Oh, I only just saw your comment after mine. Reading.

mara004: > Seems android is the problem, on Android font resolution goes through CFX_AndroidFontInfo / CFPF_SkiaFontMgr internally, bypassing the FPDF_SYSFONTINFO callbacks entirely. I agreed with opt-in class design.

Hmm, there is also an android impl of CreateDefaultSystemFontInfo though, it seems?
Or is the point that the callbacks were not honored even if there is a default sysfontinfo?

scyyh11: Android does have `CAndroidPlatform::CreateDefaultSystemFontInfo()`, but it can return `nullptr` if `CFPF_SkiaDeviceModule::GetFontMgr()` is null, so `FPDF_GetDefaultSystemFontInfo()` would return `NULL` in that case.

When it does succeed, the returned font info is a valid `SystemFontInfoIface` that gets wrapped as `FPDF_SYSFONTINFO` through the public API path, so our `MapFont`/`GetFont` interception should work in principle. Probably need more testing before we draw a conclusion.

mara004: Ah, thank you for the explanation!
Does that mean pdfium must have been built with Skia?

scyyh11: From my understanding, no. `CFPF_SkiaFontMgr` is part of PDFium's Android font stack, not proof that the build enabled the separate `pdf_use_skia` rendering backend.

`CAndroidPlatform::CreateDefaultSystemFontInfo()` calls `CFPF_SkiaDeviceModule::GetFontMgr()`, and returns `nullptr` only if that font manager is unavailable. When it succeeds, it creates `CFX_AndroidFontInfo` on top of that Android font manager.

So `GetFontMgr() == nullptr` would indicate failure to initialize PDFium's Android font manager path, not "Skia rendering support is missing from the build".

mara004: All right. I believe it may just work then. My guess is that this is just the kind of safety-oriented error handling typical of C(++) and not an anticipated case. Similar to how we have error checks in place for pdfium's APIs but these cases don't usually happen in practice unless something went really wrong, like wrong usage or maybe a broken PDF.

mara004:
@scyyh11 FWIW, https://github.com/pypdfium2-team/pypdfium2/compare/1e9aa517aa26cadd8901dc0220e32c54f4249566...2e4c2f55ddf89bc50e0bea76fac9e5c65ea46400 adds an experimental font listing CLI.

Also, https://github.com/pypdfium2-team/pypdfium2/commit/1e9aa517aa26cadd8901dc0220e32c54f4249566 exposes pdfium's default TTF map.

mara004:
I also plan to add a test file. Figured out that Scribus has an option to disable font embedding.
Libreoffice (at least the version I have) doesn't seem able to do this.

src/pypdfium2/_helpers/textpage.py
```diff
@@ -244,6 +245,33 @@ def get_textobj(self, index):
...
+                PDFium resolved it to. Empty dict if all fonts are embedded.
+        """
+        substitutions = {}
+        for obj in self.page.get_objects(filter=[pdfium_c.FPDF_PAGEOBJ_TEXT]):
+            font = obj.get_font()
+            if font.is_embedded():
+                continue
+            base_name = font.get_base_name()
+            family_name = font.get_family_name()
+            substitutions.setdefault(base_name, set()).add(family_name)
```
mara004, line 271: I came across a similar problem elsewhere. FWIW defaultdict may be a good candidate, see the examples: https://docs.python.org/3/library/collections.html#defaultdict-examples

This avoids constructing a new set (or list or whatever) when there already is a value.

mara004:
I am not 100% sure or don't remember correctly, but in retrospect I think you have used AI, right?

In any case, AI issues and PRs are now banned from this project:
https://github.com/pypdfium2-team/pypdfium2/discussions/449, https://github.com/pypdfium2-team/pypdfium2#no-ai-issues-and-prs-policy
The issue and PR templates have been updated accordingly. Just so you know.

mara004:
Ah yes, https://github.com/pypdfium2-team/pypdfium2/pull/415#issuecomment-4189996792 acknowledged AI use. I am getting forgetful.

PR #414, opened 2026-03-28
FPDF_SYSFONTINFO wrapper

mara004:
CC #413


mara004:
Apologies for mixing the move of `pypdfium2._cli` to `pypdfium2_cli` into this change.
The issue with logging not working properly during the init phase just got really vexing while working on the sysfont callback system, so I wanted to address this once and for all.

PR #417, opened 2026-04-04
FPDF_SYSFONTINFO wrapper + other progress

mara004:
Supersedes #414, see https://github.com/pypdfium2-team/pypdfium2/pull/414#issuecomment-4185919080

### DIFF
--- a/docs/devel/changelog_staging.md
+++ b/docs/devel/changelog_staging.md
@@ -6,6 +6,7 @@
 # Changelog for next release
 - `PdfDocument`: When encoding input filepath to `UTF-8`, use the `surrogateescape` error handler (except on Windows).
   This fixes loading some garbled filenames, where a default `.encode("utf-8")` call would raise `UnicodeEncodeError`. Thanks to Filipe Litaiff for the report.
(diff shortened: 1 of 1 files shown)


src/pypdfium2/_helpers/sysfontinfo.py
```diff
@@ -0,0 +1,226 @@
...
+        
+        callbacks = {n: getattr(self, n) for n in _CallbackNames}
+        if self.version != 1:  # as per docs
+            del callbacks["EnumFonts"]
+        pdfium_i.set_callbacks(self.raw, **callbacks)
+    
+    
+    def _forward_default_callbacks(self):
+        # for any callbacks that were not re-implemented, we forward from default to avoid needless python function calls
+        reference_class = type(self)  # or really just PdfSysfontBase?
```
mara004, line 128: @scyyh11 Do you have an opinion here whether this should be `type(self)` or just `PdfSysfontBase`?
I'm not entirely sure, but tentatively I feel like `type(self)`.

scyyh11: I think `type(self)` is better, `PdfSysfontBase` would only compare against base fallbacks and miss inherited overrides.

PR #418, opened 2026-04-05
Simplify output string handling

mara004:
Whether this is really more performant though, I haven't benchmarked.

### DIFF
--- a/src/pypdfium2/_helpers/textpage.py
+++ b/src/pypdfium2/_helpers/textpage.py
@@ -5,6 +5,7 @@ __all__ = ("PdfTextPage", "PdfTextSearcher")
 
 import ctypes
 import logging
+from codecs import decode
 import pypdfium2.raw as pdfium_c
 import pypdfium2.internal as pdfium_i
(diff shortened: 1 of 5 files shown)


src/pypdfium2/_helpers/textpage.py
```diff
@@ -74,7 +75,8 @@ def get_text_bounded(self, left=None, bottom=None, right=None, top=None, errors=
         buffer_ptr = ctypes.cast(buffer, ctypes.POINTER(ctypes.c_ushort))
         pdfium_c.FPDFText_GetBoundedText(*args, buffer_ptr, n_chars)
         
-        return buffer.raw.decode("utf-16-le", errors=errors)
+        # alternatively: buffer.raw or bytes(buffer) .decode(...)
+        return decode(buffer, "utf-16-le", errors=errors)
```
mara004, line 79: Not sure which one is better here. Tentatively, this feels a bit more direct, though I don't know (yet) what `codecs.decode()` does under the hood.

mara004:
Follow-up PR #425 returns to the memoryview approach after all.

PR #424, opened 2026-04-14
docs: use proper type shapes

mara004:


mara004:
Though, whether this really is an improvement is questionable enough. I'm guessing we would do better to just spell it out after all, except perhaps for the one complex shape `tuple[(tuple[float, float],) * 4]`.
Or better get rid of the `tuple[typeA, typeB, ...]` notation and use `(typeA, typeB, ...)` which is more readable.

mara004:
I'll close this attempt, but it's a topic to keep in mind. Might commit something directly later.
The main goals, I think, are consistency and readability. The current state is inconsistent.
Plus, `tuple[type * count]` isn't strictly valid, so `type` fails to highlight.

PR #425, opened 2026-04-14
Strings again

mara004:


mara004:
Even slicing the buffer directly still implies a copy (as the bytes object does not reflect changes to the buffer, naturally, since it is immutable). The memoryview approach taken here avoids that copy.

src/pypdfium2/_helpers/attachment.py
```diff
@@ -46,10 +48,10 @@ def get_name(self):
             str: Name of the attachment.
         """
         n_bytes = pdfium_c.FPDFAttachment_GetName(self, None, 0)
-        buffer = ctypes.create_string_buffer(n_bytes)
-        buffer_ptr = ctypes.cast(buffer, ctypes.POINTER(pdfium_c.FPDF_WCHAR))
-        pdfium_c.FPDFAttachment_GetName(self, buffer_ptr, n_bytes)
-        return buffer[:n_bytes-2].decode("utf-16-le")
+        n_units = -(n_bytes // -FPDF_WCHAR_size)  # ceildiv
+        buffer = (pdfium_c.FPDF_WCHAR * n_units)()
```
mara004, line 52: I'm guessing this approach is indeed better. I'm clearly not an expert on the matter, but creating a c_char array and casting (I think) raises alignment questions.

mara004:
Did some experimentation with timeit. `decode(memoryview(buffer)[:n_units-1], "utf-16-le")` does not perform worse than the previous approach, and might perform slightly better.

mara004:
Though I wonder if there isn't a better way of chopping off the NUL terminator without the need to create an intermediate memoryview object? On the other hand, I couldn't find one.

mara004:
I wonder how chopping off the NUL terminator from the python string (i.e. final `[:-1]`) would compare?

mara004:
Ok, no, did some reading. [String slicing makes a copy in CPython.](https://stackoverflow.com/a/64871976/15547292)

mara004:
Another thing worth trying may be if constructing another ctypes view (`buf_view = (c_ushort * size).from_buffer(buf)`) *might* be faster than slicing a memoryview?

mara004:
Okay, tried it out. memoryview seems faster.

PR #426, opened 2026-04-25
Autoclose machinery improvements, objects tracker, new helpers

mara004:
This would basically be a pre-requisite for destroying and re-initializing the library within a session.


src/pypdfium2/internal/bases.py
```diff
@@ -69,6 +73,7 @@ def __init__(self, close_func, *args, obj=None, needs_free=True, **kwargs):
         self._ex_kwargs = kwargs
         self._autoclose_state = pypdfium2_cfg._Mutable(_STATE.AUTO)
         self._uuid = uuid.uuid4() if pypdfium2_cfg.DEBUG_AUTOCLOSE else None
+        self._wref_to_self = weakref.ref(self)
```
mara004: todo: turn this into a cached property?

src/pypdfium2/internal/bases.py
```diff
@@ -69,6 +73,7 @@ def __init__(self, close_func, *args, obj=None, needs_free=True, **kwargs):
         self._ex_kwargs = kwargs
         self._autoclose_state = pypdfium2_cfg._Mutable(_STATE.AUTO)
         self._uuid = uuid.uuid4() if pypdfium2_cfg.DEBUG_AUTOCLOSE else None
```
mara004: (access DEBUG_AUTOCLOSE directly?)

mara004:
Also need to handle those objects that need closing but do not inherit from AutoCloseable, there are a few.

src/pypdfium2/_helpers/document.py
```diff
@@ -485,7 +485,7 @@ def page_as_xobject(self, index, dest_pdf):
         if not raw_xobject:
             raise PdfiumError(f"Failed to capture page at index {index} as FPDF_XOBJECT.")
         xobject = PdfXObject(raw=raw_xobject, pdf=dest_pdf)
-        self._add_kid(xobject)
+        dest_pdf._add_kid(xobject)
```
mara004, line 488: need to clarify with the pdfium docs but the inconsistency here is evident

mara004: yeah should be closed with the target, not with the creator.
the question is just, does the source pdf have to be kept alive? so far our code does not...

mara004: raised https://groups.google.com/g/pdfium/c/HYRoFBMG1l4 for this

mara004:
I'd say this is ready for main.

PR #428, opened 2026-05-05
build_native: launch an attempt to update pdfium

mara004:


mara004:
Ok, so GCC build mode basically already works.

Clang will be more of a problem, due to the many patches. Same goes for Android because it needs manual testing (no native CI available) and I don't currently have the device.

And something will have to be done about GN, either see if we can procure newer versions on all CI targets, or patch the sources to restore compatibility.

mara004:
Updating the legacy GN patch turned out easy enough.

mara004:
TODO update remaining patches:
- no_libclang_rt, clang_on_musl
- bigendian and bigendian_test
- android_build
- abseil_ppc64le_musl (probably can be dropped)

Update: all done. bigendian, bigendian_test and android_build still applied cleanly.

mara004:
Will want to update REUSE.toml to reflect the changes in patches.

mara004:
Triggered another test run:
https://github.com/pypdfium2-team/pypdfium2/actions/runs/25409583134
Should all be good now hopefully.

mara004:
For anyone wondering about line count statistics, you'll basically want to subtract some 905 lines for the libpng ppc64 VSX intrinsics. So the current `+1276 -310` would become mere `+371 - 310`, i.e. `+61`

mara004:
> Should all be good now hopefully.

Ah, had an oversight with loongarch, our CI targets identify as loongarch64 not just loong64.
New test runs:
https://github.com/pypdfium2-team/pypdfium2/actions/runs/25410210083
https://github.com/pypdfium2-team/pypdfium2/actions/runs/25410240384

PR #429, opened 2026-05-06
Let's see if we could use vendored libc++ with static clang (on glibc)

mara004:
Initial test run: https://github.com/pypdfium2-team/pypdfium2/actions/runs/25431818065


mara004:
Post-merge test run: https://github.com/pypdfium2-team/pypdfium2/actions/runs/25434058859

PR #430, opened 2026-05-06
workflows: jobs deduplication approach

mara004:


.github/workflows/sbuild_one_native.yaml
```diff
@@ -46,6 +49,9 @@ on:
       test_pdfium:
         type: boolean
         default: true
+      condition:
+        type: boolean
+        default: true
```
mara004: Bit nasty to have this in the workflow_dispatch as well, but I'm not sure what we could do about this. Not yet, anyway.

mara004:
I like this approach, and I think we should update the other workflows likewise.
This affects sbuild.yaml and cibw.yaml, as far as I can see.

.github/workflows/sbuild_one_native.yaml
mara004: Would be interesting to check what is the fallback value of `inputs.condition` if we omit it here and make a workflow dispatch.
In case it is an empty string, we could just check for that below, e.g.
`if: ${{ inputs.condition == '' || inputs.condition }}`.
However if it were false that would be impossible to distinguish from an intentional false.

mara004: > However if it were false that would be impossible to distinguish from an intentional false.

Though in that case we could use an inversion, e.g. rename `condition` to `skip` then check
`if: ${{ !inputs.skip }}`

mara004:
Been a bit unwell while writing this. Need to review this again later for possible oversights.

mara004:
OK, test runs are all green, except the Linux arm64 native toolchained test which had already been broken before this by the build_native update.

PR #431, opened 2026-05-09
Deduplicate workflow inputs, migrate from `benc-uk/workflow-dispatch` to reusable workflow call

mara004:
Unfortunately, GH YAML anchors are weak (due to lack of support for merge keys), see this excellent blog post:
https://frenck.dev/github-actions-yaml-anchors-aliases-merge-keys/

but still, it might be acceptable for our purposes.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -11,6 +11,3 @@ updates:
     schedule:
       # Check for updates to GitHub Actions every week
       interval: "weekly"
-    ignore:
-      - dependency-name: "benc-uk/workflow-dispatch"
-        versions: ["121"]  # presumably an upstream accident
--- a/.github/workflows/cibw.yaml
+++ b/.github/workflows/cibw.yaml
@@ -6,33 +6,33 @@ on:
   workflow_dispatch:
     inputs:
       cibw_py_ver:
-        default: 'cp314'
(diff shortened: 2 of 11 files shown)


mara004:
todo: mark more inputs as required

mara004:
Deduplication is probably worth the downsides to us (not having `choice` type, `condition` being included in workflow_dispatch).

Also, sharing workflow inputs may open the door to get rid of the workflow-dispatch action and migrate to reusable workflows altogether.

mara004:
This would be more of what I'd been looking for:
https://devpane.tools/yaml/yaml-anchors-aliases
but as said, it does not work on GH actions...

mara004:
Got an obscure error:
```
Error: Can't find 'action.yml', 'action.yaml' or 'Dockerfile' under '/home/runner/work/pypdfium2/pypdfium2/.github/workflows/conda.yaml'. Did you forget to run actions/checkout before running your local action?
```

mara004:
It might be because it's a step not a job? I'll try changing to a jobs, that'll be cleaner in any case.

*Edit: OK, yeah, that was the cause*

mara004:
I like this change, but maybe should re-review carefully tomorrow before merging...

PR #443, opened 2026-06-09
Setup refactor, preparations for auto-tagging

mara004:


mara004:
If we want this used in packaging, there is still a long way to go, because we'll need to split main.yaml's build phase across native hosts.

mara004:
I'll go ahead and merge this, let's see how it goes.

PR #445, opened 2026-06-10
fix: memleak when a page is retrieved from a document with an active form environment and objects are auto-closed

### LINKED ISSUE #444: Memory leak when a page is retrieved from a document with an active form environment and objects are auto-closed
### Checklist

- [x] I confirm to be using an **official package** of `pypdfium2` from `PyPI` or `GitHub/pypdfium2-team` (unmodified sources).

### Description

# Post-mortem
The following discussion had some back-and-forth (measuring runtimes with apparently unclean test environments, running into several dead ends, ..).
Before digging into details, consider reading [the summary in the end](https://github.com/pypdfium2-team/pypdfium2/issues/444#issuecomment-4688667782).

# Memory leak when a page is retrieved from a document with an active form environment and objects are auto-closed (garbage-collected)

### Environment
- pypdfium2 **5.9.0** (pdfium **150.0.7869.0**), pip wheel
- CPython 3.14, Linux x86_64

### Summary
When `init_forms()` has been called and a page is retrieved from the document, relying on automatic close (garbage collection) instead of calling `close()` leaks memory steadily (~18 kB/iteration in the repro, unbounded). Explicitly closing the **document** prevents the leak; closing only the **page** does not.

This minimal example was distilled after trying to find the reason for the following errors (various Python versions, `pypdfium 5.9.0`):
```
The following objects are still open and will now be closed: ***<PdfDocument 0x7f27daba1fd0 from '***/form_field.pdf'>, <PdfFormEnv 0x7f27b253de80>***
```
or this (Python `3.12`, `pypdfium 5.6.0`):
```
Cannot close object; pdfium library is destroyed. This may cause a memory leak.
```

### Minimal reproduction
Self-contained (generates a tiny single-field AcroForm PDF:

```python
import resource
import sys
import pypdfium2 as pdfium


def build_form_pdf(path):
    objs = [
        b"1 0 obj<</Type/Catalog/Pages 2 0 R/AcroForm<</Fields[4 0 R]>>>>endobj\n",
        b"2 0 obj<</Type/Pages/Kids[3 0 R]/Count 1>>endobj\n",
        b"3 0 obj<</Type/Page/Parent 2 0 R/MediaBox[0 0 200 100]/Annots[4 0 R]>>endobj\n",
        b"4 0 obj<</Type/Annot/Subtype/Widget/FT/Tx/Rect[10 10 190 40]/T(field1)/V(hello)>>endobj\n",
    ]
    data = b"%PDF-1.4\n"
    offsets = []
    for o in objs:
        offsets.append(len(data))
        data += o
    xref_pos = len(data)
    data += b"xref\n0 5\n0000000000 65535 f \n"
    for off in offsets:
        data += ("%010d 00000 n \n" % off).encode()
    data += b"trailer<</Size 5/Root 1 0 R>>\nstartxref\n%d\n%%%%EOF\n" % xref_pos
    with open(path, "wb") as f:
        f.write(data)


def rss_kb():
    return resource.getrusage(resource.RUSAGE_SELF).ru_maxrss


close = "--close" in sys.argv
pdf_path = "/tmp/pypdfium2_form.pdf"
build_form_pdf(pdf_path)
print(f"pypdfium2 {pdfium.version.PYPDFIUM_INFO}, pdfium {pdfium.version.PDFIUM_INFO}")

start = rss_kb()
for _ in range(3000):
    doc = pdfium.PdfDocument(pdf_path)
    if close:
        try:
            doc.init_forms()
            page = doc[0]
            page.close()
        finally:
            doc.close()
    else:
        doc.init_forms()
        page = doc[0]  # rely on GC, no explicit close => leaks

print(f"{'CLOSE  ' if close else 'NOCLOSE'}: maxrss +{rss_kb() - start} kB over 3000 iterations")
```

### Observed output
```
pypdfium2 5.9.0, pdfium 150.0.7869.0
NOCLOSE: maxrss +53328 kB over 3000 iterations
pypdfium2 5.9.0, pdfium 150.0.7869.0
CLOSE  : maxrss +0 kB over 3000 iterations
```

### Bisection of the trigger
| Scenario (no explicit close, relying on GC)                         | Leak?      |
|---------------------------------------------------------------------|------------|
| open document only                                                  | no         |
| `init_forms()`, no page retrieved                                   | no         |
| `init_forms()` + `doc[0]` (no render)                               | **yes**    |
| `doc[0]` + `render()`, **without** `init_forms()`                   | no         |
| `init_forms()` + `doc[0]` + `render()`                              | **yes**    |
| `init_forms()` + `doc[0]` + `render()`, then `doc.close()` only     | no         |
| `init_forms()` + `doc[0]` + `render()`, then `page.close()` only    | **yes**    |
| `init_forms()` + `doc[0]` + `render()`, then `page.close()` + `doc.close()` | no |

So:
- The leak requires **both** `init_forms()` **and** retrieving a page. Rendering is not required.
- Without `init_forms()`, retrieving + rendering a page and relying on GC does **not** leak.
- Explicitly closing the **document** prevents it; closing only the **page** does not.

### Suspected cause
`PdfDocument.formenv` and `PdfFormEnv.pdf` reference each other, so a document with an initialized form environment forms a reference cycle and is reclaimed by the *cyclic* collector rather than by reference counting. During cyclic collection, `weakref.finalize` callbacks are not guaranteed to run in child-before-parent order, and the form-fill teardown (`FPDFDOC_ExitFormFillEnvironment`) reached via the finalizer path in this scenario does not release the page-associated form-fill memory that an explicit, ordered `close()` does.

### Workaround
Always close the document explicitly (e.g. in a `finally:` / context manager) when `init_forms()` is used:

```python
doc = pdfium.PdfDocument(path)
try:
    doc.init_forms()
    page = doc[0]
    try:
        ...  # render etc.
    finally:
        page.close()
finally:
    doc.close()
```

## Additional information
Written and distilled by assistance of Claude.

### Install Info

```shell
Python 3.14.0 (main, Oct 14 2025, 21:27:55) [Clang 20.1.4 ]
Linux-6.17.0-1025-oem-x86_64-with-glibc2.39
pypdfium2 5.9.0
pdfium 150.0.7869.0
```

noxthot:
## Description

Fixes #444 — memory leak when a document with an active form environment is reclaimed by GC instead of being closed via `close()` explicitly (~18 kB per document/page in the linked repro, unbounded growth).

_Note: Assisted by Claude._

### Root cause

Two layers, both in the autoclose machinery's interaction with the global `weakref.finalize` registry:

1. **Immortal reference cycle**: `PdfFormEnv`'s finalizer state (the finalize `args` and `_FinalizerOwner.parent`) held strong references to the parent `PdfDocument`, while `pdf.formenv` points back at the formenv. Since the finalize registry is a GC root and its entry is only released when the finalizer runs, the document <-> formenv pair was permanently reachable and **never collected at all** — the exact anti-pattern the `weakref.finalize` docs warn about (and which `_close_template`'s comment already quotes). Verified empirically: after a forced `gc.collect()`, all documents from a loop were still alive. With a page loaded, each immortal document additionally retains its parsed page cache, hence the leak size.

2. **GC generation promotion**: even with (1) fixed, `PdfPage`'s finalizer args held the formenv strongly, so the document/formenv pair only became garbage one GC pass *after* the page's collection — by which point CPython (tested on 3.13) had usually promoted them to an older generation that effectively never gets collected in a steady-state workload. Still unbounded growth in practice.

### Fix

- New `AutoCloseable._WEAK_PARENT` opt-in (`internal/bases.py`): the finalizer state holds the parent via weakref. Set for `PdfFormEnv` and `PdfPage`.
- Holding parents weakly forfeits the guaranteed child-before-parent finalization order when document, formenv and pages are reclaimed in the same GC run. To compensate, mutable raw-handle holders (`PdfDocument._formenv_holder`, `._page_holders`) are shared between the finalizers of document, formenv and pages: whichever close function runs first takes over pending teardown work in due order (`FORM_OnBeforeClosePage` -> `FPDF_ClosePage` -> `FPDFDOC_ExitFormFillEnvironment` -> `FPDF_CloseDocument`), and the remaining calls degrade to no-ops.
- Liveness guarantees for *live* helper objects are unchanged (instance attributes like `page.pdf`, `page.formenv`, `formenv.pdf` remain strong refs), and explicitly `close()`d objects behave exactly as before (kids-walk order preserved).

### Verification

- Repro from #444: `NOCLOSE` went from **+53 MB (unbounded)** to bounded behavior matching the no-forms baseline over 20k iterations; live object counts oscillate as GC reclaims them instead of growing monotonically.
- Full test suite passes. New tests: GC collectability regression test, all six manual orderings of the three close functions (takeover logic), explicit `formenv.close()` + re-`init_forms()`, page GC while document/formenv stay alive and usable.
- Worst-case ordering was also observed live via `DEBUG_AUTOCLOSE`: in a real GC group death the *document's* finalizer fired first and correctly performed the whole teardown.

## Checklist

</details>
<details close><summary>Setup</summary>

</details>

- [x] This PR changes other things, namely: internal autoclose machinery (`internal/bases.py`: `_WEAK_PARENT` opt-in)

</details>

### DIFF
--- a/docs/devel/changelog_staging.md
+++ b/docs/devel/changelog_staging.md
@@ -4,6 +4,7 @@
 <!-- List character: dash (-) -->
 
 # Changelog for next release
+- Fixed a memory leak with form environments: `PdfDocument.formenv` and the formenv's finalizer state referenced each other, rooting the document <-> formenv reference cycle in the global `weakref.finalize` registry, so that it could never be garbage collected. Also, `PdfPage`'s finalizer state held strong references to formenv/document, which (due to GC generation promotion) could delay their collection indefinitely. Finalizer states now hold parents weakly; correct child-before-parent teardown order in the GC case is ensured by sharing mutable raw-handle holders between the finalizers of document, formenv and pages, so whichever runs first takes over pending teardown work in due order. Explicit `close()` calls behave as before, and remain recommended for deterministic resource release.
 - Updated `build_toolchained.py` from PDFium `7191` to `7880`. Despite the gap, this turned out fairly straightforward.
 - Updated `build_native.py` from PDFium `7841` to `7880` (see below for more info).
   Added patch to fix a build system issue introduced upstream.
--- a/src/pypdfium2/internal/bases.py
+++ b/src/pypdfium2/internal/bases.py
@@ -76,8 +76,13 @@ def _close_template(info, owner):
         return
     
     assert info.state != _STATE.INVALID
-    
+
     parent = owner.parent
+
+    if isinstance(parent, weakref.ref):
+        # _WEAK_PARENT - may be None if the parent is dead or trash in the same GC run
+        parent = parent()
+
     if parent is not None:
         assert not parent._tree_closed()
         if info.tracked:
@@ -103,7 +108,12 @@ class AutoCastable:
 
 
 class AutoCloseable (AutoCastable):
-    
+
+    # If True, the finalizer state holds the parent via weakref rather than a strong reference.
+    # This must be set for objects the parent strongly references back (e.g. document <-> formenv): otherwise, the global finalizer registry would root the resulting reference cycle, so that it could never be garbage collected.
+    # The price is that child-before-parent finalization order is not guaranteed anymore when both objects are reclaimed in the same GC run, so the parent's close function must be able to handle the child's cleanup if it runs first (cf. PdfDocument/PdfFormEnv._close_impl).
+    _WEAK_PARENT = False
+
     def __init__(self, close_func, *args, obj=None, needs_free=True, tracked=True, **kwargs):
         
         # proactively prevent accidental double initialization
@@ -133,7 +143,12 @@ class AutoCloseable (AutoCastable):
         own_type = type(self)
         # note, this captures the object's parent, repr and so on at finalizer installation time
         # in case they ever change, we'd have to store the owner in an attribute and update it
-        owner = _FinalizerOwner(self.raw, self.parent, self._wref_to_self, own_type, repr(self))
+        parent = self.parent
+        
+        if self._WEAK_PARENT and parent is not None:
+            parent = parent._wref_to_self
+            
+        owner = _FinalizerOwner(self.raw, parent, self._wref_to_self, own_type, repr(self))
         self._finalizer = weakref.finalize(self._fin_obj, _close_template, self._fin_info, owner)
         ObjectTracker[own_type].add(self._wref_to_self)
     
--- a/src/pypdfium2/_helpers/document.py
+++ b/src/pypdfium2/_helpers/document.py
@@ -68,16 +68,18 @@ class PdfDocument (pdfium_i.AutoCloseable):
         self._autoclose = autoclose
         self._data_holder = []
         self._data_closer = []
+        self._formenv_holder = []
+        self._page_holders = []
         self.formenv = None
-        
+
         if isinstance(self._input, pdfium_c.FPDF_DOCUMENT):
             self.raw = self._input
         else:
             self.raw, to_hold, to_close = _open_pdf(self._input, self._password, self._autoclose)
             self._data_holder += to_hold
             self._data_closer += to_close
-        
-        super().__init__(PdfDocument._close_impl, self._data_holder, self._data_closer, tracked=False)
+
+        super().__init__(PdfDocument._close_impl, self._data_holder, self._data_closer, self._formenv_holder, self._page_holders, tracked=False)
     
     
     # Support using PdfDocument in a with-block
@@ -108,7 +110,21 @@ class PdfDocument (pdfium_i.AutoCloseable):
     
     
     @staticmethod
-    def _close_impl(raw, data_holder, data_closer):
+    def _close_impl(raw, data_holder, data_closer, formenv_holder, page_holders):
+        # Pages or formenv may not have been closed in due order before the document: when document, formenv and pages are reclaimed in the same GC run, finalizer order is not guaranteed. Take over any pending teardown now, before closing the document.
+        for holder in page_holders:
+            if holder:
+                if formenv_holder:
+                    pdfium_c.FORM_OnBeforeClosePage(holder[0], formenv_holder[0])
+                pdfium_c.FPDF_ClosePage(holder[0])
+                holder.clear()
+
+        page_holders.clear()
+        
+        if formenv_holder:
+            pdfium_c.FPDFDOC_ExitFormFillEnvironment(formenv_holder[0])
+            formenv_holder.clear()
+
         pdfium_c.FPDF_CloseDocument(raw)
         for data in data_holder:
             id(data)
@@ -173,6 +189,9 @@ class PdfDocument (pdfium_i.AutoCloseable):
         raw = pdfium_c.FPDFDOC_InitFormFillEnvironment(self, config)
         if not raw:
             raise PdfiumError(f"Initializing form env failed for document {self}.")
+        
+        assert not self._formenv_holder
+        self._formenv_holder += (raw, config)
         self.formenv = PdfFormEnv(raw, self, config)
         self._add_kid(self.formenv)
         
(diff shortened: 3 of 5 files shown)


mara004:
AI mess and definitely not mergable, I cannot afford to waste time on this. Consider creating a human-only PR.

docs/devel/changelog_staging.md
```diff
@@ -4,6 +4,7 @@
 <!-- List character: dash (-) -->
 
 # Changelog for next release
+- Fixed a memory leak with form environments: `PdfDocument.formenv` and the formenv's finalizer state referenced each other, rooting the document <-> formenv reference cycle in the global `weakref.finalize` registry, so that it could never be garbage collected. Also, `PdfPage`'s finalizer state held strong references to formenv/document, which (due to GC generation promotion) could delay their collection indefinitely. Finalizer states now hold parents weakly; correct child-before-parent teardown order in the GC case is ensured by sharing mutable raw-handle holders between the finalizers of document, formenv and pages, so whichever runs first takes over pending teardown work in due order. Explicit `close()` calls behave as before, and remain recommended for deterministic resource release.
```
mara004, line 7: This changelog entry makes somewhat more sense to me than the rest.
I will give this some thought. I can see it might be problematic that PdfFormEnv's close impl references the parent pdf.

noxthot:
Just to note it also here:
Following the discussions in #444, I found that the "bigger" problem appears to be unrelated to PDFs with forms. See this [post](https://github.com/pypdfium2-team/pypdfium2/issues/444#issuecomment-4670863593) for more information.

Yet, this PR also seems to solve parts of the bigger problem as noted in [this comment](https://github.com/pypdfium2-team/pypdfium2/issues/444#issuecomment-4670913830)

Ofc only if I am not misunderstanding something fundamental and growing memory consumption when not closing explicitely is the expected behaviour (as also noted in the [first mentioned post](https://github.com/pypdfium2-team/pypdfium2/issues/444#issuecomment-4670863593))

src/pypdfium2/internal/bases.py
```diff
@@ -76,8 +76,13 @@ def _close_template(info, owner):
         return
     
     assert info.state != _STATE.INVALID
-    
+
     parent = owner.parent
+
+    if isinstance(parent, weakref.ref):
+        # _WEAK_PARENT - may be None if the parent is dead or trash in the same GC run
+        parent = parent()
```
mara004, line 84: The concept of a weak parent is arrant nonsense. Your clanker clearly had none of the domain knowledge needed here, and has been making changes without the remotest idea of what it is actually doing.

Why is it nonsense? The API contract with PDFium requires that child objects be closed before parents, and that the parent is still alive while the child is being closed.
When the parent is gone before the child, we are already in an invalid state. (!)

Children strong-referencing their parent is correct in principle (e.g. page referencing parent pdf).
The issue stems from parent objects strong-referencing child objects when the child object is part of the autoclose machinery and in turn references the parent.

mara004: The page holders concept is also utter rubbish, but I don't want to waste more time commenting on it.

Please consider that AI can output huge amounts of garbage in split seconds of machine time while it takes humans real lifetime to untangle and explain that garbage.

That is precisely why I hate (low-quality) AI PRs.

noxthot: I am very well aware that AI can produce lots of nonsense, but again: this PR was not created by one prompt, telling some AI to fix it and going yolo, publishing a PR without looking at it myself.

I have re-iterated several times and thought to have come up with a somehat okay PR. I am neither an expert in the context of memleaks nor I have prior experience with the `pypdfium2` source code and memleak/constructor-deconstructor issues are usually harder problems (at least for me), especially when they are observed in "the wild" with many other processes/packages/side effects being involved.

The "very limited domain knowledge" applies to most of the one-time contributors in open-source projects. That's why PRs exist and nothing is merged without someone having that context (reviewing the PR).

Furthermore it is not like I submitted an AI-bloated mega-PR. This PR adds +177 lines, more than half of it being only tests and changelog. Another half (this is estimated) is comments. So it's about roughly +50 -20 lines that actually did something in the code and from what I can see in comparison with your PR #447: Those changes touched the "right spots"; but apparently in the wrong way and more spots than needed; which also sometimes happens when someone decides to open up a PR without AI assistance.

Having said that: I did not expect anyone to dissect every line of this code to explain why it is garbage. When you as the domain expert look at it, and immediately see that it breaks core concepts, then a simple "breaks core concept of XY" is enough information; and this info would probably also help to re-iterate and improve on this (or a new) PR.

mara004: Nevermind, sorry for the rant. I was merely a bit .. revolted by what the change did to the code combined with previous negative experience from AI-assisted PRs.

noxthot: totally understand <3

mara004:
See https://github.com/pypdfium2-team/pypdfium2/issues/444#issuecomment-4760000055

PR #454, opened 2026-06-25
Add `PdfBookmark.get_color()` API

mara004:


src/pypdfium2/_helpers/document.py
```diff
@@ -699,6 +699,18 @@ def get_count(self):
...
+        """
+        Returns:
+            tuple[float,float,float] | None: The bookmark's RGB color, otherwise None if the bookmark does not specify a color or on failure.
+        """
+        # requires pdfium > 7912
+        r, g, b = ctypes.c_float(), ctypes.c_float(), ctypes.c_float()
+        ok = pdfium_c.FPDFBookmark_GetColor(self, r, g, b)
+        if not ok:
+            return None
+        return r.value, g.value, b.value
```
mara004, line 712: need to acknowledge that it is 0-1 (float), not 0-255 (int)

mara004:
I reckon we could drop `toc_colored.pdf` after all since `toc_viewmodes.pdf` already happens to have colors.

PR #465, opened 2026-07-19
Python 3.6 compat fixes (runtime & setup)

mara004:


src/pypdfium2_stl/shared/stl.py
```diff
@@ -0,0 +1,44 @@
+# SPDX-FileCopyrightText: 2026 geisserml <geisserml@gmail.com>
+# SPDX-License-Identifier: Apache-2.0 OR BSD-3-Clause
+
+class cached_property:
+    """
+    Custom cached property implementation.
+    
+    Powered by the descriptor protocol. Zero overhead after the property has been loaded.
+    Similar to :class:`functools.cached_property`, but cleaner and more backward compatible.
```
mara004: (Minor FIXME: for some reason `functools.cached_property` fails to link with intersphinx. But it is definitely a class, isn't it?)

mara004: Aah, no, it is a class, but also a decorator, so sphinx wants us to use the `:deco:` directive...

mara004: nvm, deco still failed to link. `:obj:` actually works.
https://www.sphinx-doc.org/en/master/usage/domains/python.html#role-py-obj

PR #478, opened 2026-08-10
Improve `pip install` integration (again)

mara004:


mara004:
The next step will be to make the gn version configurable.
That's what I originally wanted to do, but then found reworking the pip install integration would be a prerequisite.

mara004:
58928321eaf8ab8c9468960052b6be898d5b4b00 fixes an oversight here

PR #481, opened 2026-08-19
Drop conda packaging, module splitting and `system-generate` target

mara004:
See https://github.com/pypdfium2-team/pypdfium2/discussions/480 / https://github.com/mindee/doctr/discussions/2127 and https://github.com/mindee/doctr/issues/113#issuecomment-5339953835 ff.


mara004:
Need to remember to remove and revoke the secret API token as well.

mara004:
This will be rather quicker than originally intended, but I think I will move on with this, since I'm committed to the decision, and have some changes on my mind that I don't want to run into merge conflicts with.

README.md
```diff
@@ -332,99 +328,34 @@ As it is hard to keep up with constantly evolving setup code, it is possible thi
...
 
 [^platform_ids]: Intended for packaging, so that wheels can be crafted for any platform without access to a native host.
 
 
-### From Conda
+### If you're using conda
 
-> [!WARNING]
-> **Beware:** Any conda packages/recipes of pypdfium2 or pdfium-binaries that might be provided by other distributors, including `anaconda/main` or `conda-forge` default channels, are [unofficial](#unofficial-packages).
+...
```
mara004: This still needs to be addressed before merging though. Either we need to write something here, or remove the section.

mara004: Ok, I feel like the changelog and linked discussion ticket now give a good enough explanation of what to do when you're using conda: Install pypdfium2 from PyPI anyway (e.g. via pip), or if it's a package recipe, just don't list pypdfium2 as a conda dependency and ask users to post-install it via pip instead ~~or consider using the new conda-pypi bridge~~.

PR #483, opened 2026-09-02
Rework ctypesgen integration

### LINKED ISSUE #482: Clone ctypesgen into pypdfium2 and bundle in sdists
Pulling ctypesgen into pypdfium2 as a git submodule and bundling its `src/` in the sdist would fix the issue with our sdists being non‑hermetic, i.e. dependent on source from github instead of being packages only. This is the only valid issue I ever saw in https://github.com/pypdfium2-team/ctypesgen/issues/17.

As a positive side effect, we would no longer have to monkeypatch `pyproject.toml` to pin ctypesgen when building an sdist, and we could (and probably will) remove ctypesgen's setup code entirely, for I never really maintained it and it just adds complexity (with versioning etc.) and potentially incompatibilities that we don't need. Users and embedders can add ctypegsen's `src/` to `PYTHONPATH` without any setup tooling and invoke via `python -m ctypesgen`, or add an entrypoint `bin/` to `PATH`.

mara004:
Coupled with https://github.com/pypdfium2-team/ctypesgen/pull/47

### DIFF
--- a/setupsrc/base.py
+++ b/setupsrc/base.py
@@ -663,7 +663,7 @@ class _LazyClass:
         import ctypesgen
         import ctypesgen.__main__
         if not ctypesgen.__version__.split(" ")[0] == "pypdfium2-ctypesgen":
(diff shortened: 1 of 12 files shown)


mara004:
Hmm maybe we should use `git subtree`? Let's see...

setupsrc/base.py
```diff
@@ -632,14 +635,45 @@ def tmp_cwd_context(tmp_cwd):
 
 
 CTG_LIBPATTERN = "{prefix}{name}.{suffix}"
+_CTG_BRANCH = "remove_setup"  # XXX "pypdfium2"
```
mara004: Reminder: This must be changed back on before merge

mara004: done

PR #484, opened 2026-09-05
Integrate new API `FPDFBookmark_GetStyle()`

mara004:


mara004:
... should yet add some proper API test cases

Also, something will have to be done about test suite compatibility with older pdfium versions.
Same goes for the new attachment APIs. Perhaps just updating the build scripts altogether would be best.

